PR #27, opened 2023-02-23
Add basic Identity auth flow

malcolmbaig:
### Context

As a qualified teacher
I want to sign in to Access your teaching certificates using Teacher ID
So that I can use a common sign in mechanism across all teacher services

Add a basic implementation of User sign in via Get an identity. See commits for a summary of constituent parts.

### Guidance to review
Can be tested locally with Identity credentials in local .env.development file (I can send these in DM).

This PR outlines the implementation, things that need some more thought and will follow:

- How and what to test
- How to handle expired tokens

### Link to Trello card
https://trello.com/c/stXI0dzN/664-sign-in-to-quals-service-using-teacher-id

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/users/omniauth_callbacks_controller.rb
+++ b/app/controllers/users/omniauth_callbacks_controller.rb
@@ -1,0 +1,30 @@
+# frozen_string_literal: true
+
+class Users::OmniauthCallbacksController < Devise::OmniauthCallbacksController
+  # You should configure your model like this:
+  # devise :omniauthable, omniauth_providers: [:twitter]
+
+  # You should also create an action method in this controller like this:
+  # def twitter
+  # end
+
+  # More info at:
+  # https://github.com/heartcombo/devise#omniauth
+
+  # GET|POST /resource/auth/twitter
+  # def passthru
+  #   super
+  # end
+
+  # GET|POST /users/auth/twitter/callback
+  # def failure
+  #   super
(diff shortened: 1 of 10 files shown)


app/controllers/users/omniauth_callbacks_controller.rb
```diff
@@ -0,0 +1,30 @@
...
+  # def failure
+  #   super
+  # end
+
+  # protected
+
+  # The path used when OmniAuth fails
+  # def after_omniauth_failure_path_for(scope)
+  #   super(scope)
+  # end
```
felixclack, line 29: Do we need these comments?

spec/models/user_spec.rb
```diff
@@ -0,0 +1,4 @@
+require "rails_helper"
+
+RSpec.describe User, type: :model do
```
felixclack, line 3: Would be good to add tests for the `from_identity` method.

felixclack (approved):

app/controllers/users/omniauth_callbacks_controller.rb
malcolmbaig: Not essential, but they're useful hints about other functionality in this class that we'll likely use in the near term.

PR #29, opened 2023-02-27
Add OkComputer for a health check endpoint

felixclack:
We want a way to ping an endpoint for a system status message.

This allows us to automate monitoring of the service status.

OkComputer has been used in other projects (Find and Refer) and this
implementation follows the configuration there.

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased main
- [ ] Cleaned commit history


malcolmbaig (approved):
Spec failure looks like an unmet Notify dependency.

PR #30, opened 2023-02-27
Add Sidekiq

felixclack:
As per the initial ADR, we have opted to use Sidekiq as the background
worker system.

We would like to send emails via a background process.

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased main
- [ ] Cleaned commit history

### DIFF
--- a/config/initializers/sidekiq.rb
+++ b/config/initializers/sidekiq.rb
@@ -1,0 +1,9 @@
+# frozen_string_literal: true
+
+Sidekiq.configure_server do |config|
+  config.redis = { url: ENV.fetch("REDIS_URL", "redis://localhost:6379/1") }
+end
+
+Sidekiq.configure_client do |config|
+  config.redis = { url: ENV.fetch("REDIS_URL", "redis://localhost:6379/1") }
+end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -1,3 +1,5 @@
+require "sidekiq/web"
+
 Rails.application.routes.draw do
   root to: "pages#home"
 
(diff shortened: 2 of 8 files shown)


config/routes.rb
```diff
@@ -24,4 +26,10 @@
   get "/accessibility", to: "static#accessibility"
   get "/cookies", to: "static#cookies"
   get "/privacy", to: "static#privacy"
+
+  devise_scope :staff do
```
malcolmbaig: Nit: worth co-locating this around line 14 with the other staff / support stuff

config/initializers/sidekiq.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+Sidekiq.configure_server do |config|
+  config.redis = { url: ENV.fetch("REDIS_URL", "redis://localhost:6379/1") }
+end
+
+Sidekiq.configure_client do |config|
+  config.redis = { url: ENV.fetch("REDIS_URL", "redis://localhost:6379/1") }
```
malcolmbaig, line 8: Do we need to add REDIS_URL [here](https://github.com/DFE-Digital/access-your-teaching-qualifications/blob/main/Dockerfile#L43) if we're using a fetch in this initializer?

malcolmbaig (approved):
Couple of non-blocking comments, otherwise LG 👍🏼

config/initializers/sidekiq.rb
felixclack: Not sure about this one. I don't use Docker. Happy to add it but can't verify if it does anything.

felixclack: Looks like we don't do that in Refer. 🤷

PR #40, opened 2023-03-02
Add the UI for qualifications

felixclack:
The prototype has a [design for the qualifications screen](https://qualifications-prototype.herokuapp.com/landing-pages/v2/qualifications).

We want to recreate this in the app.

A lot of the data being displayed isn't present as it will be fetched
from the Qualifications API.

In order to allow us to build the views, I opted to stub out the data
with values from the prototype. The aim would be to replace these with
the data from the API once that connection is setup.

I upgraded the `govuk-components` dependency to the `4.0.0a2` version so
that we have access to the new summary card component.

This requires a few changes elsewhere in the views due to breaking
changes there.

Screenshot 2023-03-02 at 1 11 40 pm

### Link to Trello card

https://trello.com/c/JSVZFZ6b/672-add-qualification-list-ui

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased main
- [ ] Cleaned commit history

### DIFF
--- a/app/views/qualifications/show.html.erb
+++ b/app/views/qualifications/show.html.erb
@@ -1,0 +1,32 @@
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-full">
+    <span class="govuk-caption-xl"><%= @user.name.possessive %></span>
+    <h1 class="govuk-heading-xl">Teaching qualifications</h1>
+
+    <div class="govuk-grid-row">
+      <div class="govuk-grid-column-two-thirds-from-desktop">
+        <%= render InductionSummaryComponent.new(qualification: @induction) if @induction %>
+        <%= render QtsSummaryComponent.new(qualification: @qts) if @qts %>
+        <%= render IttSummaryComponent.new(qualification: @itt) if @itt %>
+      </div>
+
+      <div class="govuk-grid-column-one-third app__details-menu">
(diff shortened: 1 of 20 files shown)


malcolmbaig (approved):
Nice one 👌🏼 . Worth doing Mandatory qualification as well, in either this or a followup PR? ref: https://qualifications-prototype.herokuapp.com/landing-pages/v2/qualifications

app/views/qualifications/show.html.erb
```diff
@@ -0,0 +1,32 @@
...
+        </p>
+
+        <p>
+          Date of birth<br /> 
+          <strong><%= @user.date_of_birth.to_fs(:long_uk) %></strong>  
+        </p>
+
+        <p>
+          Teacher Reference number (TRN)<br /> 
+          <strong><%= @user.trn %></strong>
```
malcolmbaig, line 27: One thing to note here — looking at the [language used in the prototype](https://qualifications-prototype.herokuapp.com/landing-pages/v2/account-details), I think this info comes from the `/v3/teacher` API call rather than from the User object (which caches details from Identity).

@gpeng — does that sound about right? I can card up updating this if so.

gpeng: Yes. I think so. This should be from DQT

PR #55, opened 2023-03-13
Fetch name & TRN from DQT API

felixclack:
The user details that get returned in the authentication process aren't
guaranteed to match those stored in DQT.

We want to display the DQT data in this service.

### Link to Trello card

https://trello.com/c/fMyKdIKv/761-use-quals-api-to-populate-your-details

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/app/views/qualifications/show.html.erb
+++ b/app/views/qualifications/show.html.erb
@@ -14,7 +14,7 @@
         <h3 class="govuk-heading-m">Your details</h3>
         <p>
           Name on certificates<br /> 
-          <strong><%= @user.name.full %></strong>
+          <strong><%= [@teacher.first_name, @teacher.last_name].join(" ") %></strong>
         </p>
 
         <p>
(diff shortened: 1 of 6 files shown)


app/views/qualifications/show.html.erb
```diff
@@ -14,7 +14,7 @@
         <h3 class="govuk-heading-m">Your details</h3>
         <p>
           Name on certificates<br /> 
-          <strong><%= @user.name.full %></strong>
+          <strong><%= [@teacher.first_name, @teacher.last_name].join(" ") %></strong>
         </p>
```
malcolmbaig, line 18: I assume the user name just above the h1 should remain the one we get from Identity?

malcolmbaig (approved):

app/views/qualifications/show.html.erb
felixclack: I guess so. It's strange to have competing sources for this data.

PR #59, opened 2023-03-14
Handle expired tokens

malcolmbaig:
### Context

We use an access token to query the Quals API. This token is provided by Identity during the auth flow, and saved in the user’s session. The token has an expiry. We need to have a check in place to ensure user requests to the Quals service don’t result in an API call with an expired token.

- We check the token expiry on each request
- If expired, sign the user out and prompt them to re-authenticate

This PR also includes some improvements identified while implementing the above:
- Remove the default Rails home screen
- Simplify the Identity Omniauth strategy

See commit messages for detail.
### Link to Trello card
https://trello.com/c/TGdd619m

### Checklist

- [x] Attach to Trello card
- [ ] Rebased main
- [x] Cleaned commit history


app/controllers/qualifications_interface_controller.rb
```diff
@@ -0,0 +1,13 @@
+class QualificationsInterfaceController < ApplicationController
+  before_action :authenticate_user!
+  before_action :handle_expired_token!
+
+  def handle_expired_token!
+    redirect_to sign_out_path unless session[:identity_user_token_expiry]
+
+    if Time.zone.now > Time.zone.at(session[:identity_user_token_expiry])
```
felixclack: Just a nit picking suggestion here. I like to use the ActiveSupport helpers to make code like this more readable. `>` and `<` always confuse me.

```suggestion
    if Time.zone.at(session[:identity_user_token_expiry]).past?
```

felixclack (approved):

PR #68, opened 2023-03-17
Add NPQs to the qualifications list

felixclack:
The DQT API now returns all the NPQs associated with the teacher.

We can display this using a similar approach to the other
qualifications.

One difference is the certificates URL. As there can be multiple
certificates, we need to rely on the value returned for the specific NPQ
and pass that to the client as the download link.

### Link to Trello card

https://trello.com/c/THIzW32K/817-add-npq-summaries-from-the-api-to-the-qualifications-page

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

Screenshot 2023-03-17 at 12 25 44 pm

### DIFF
--- a/app/components/npq_summary_component.html.erb
+++ b/app/components/npq_summary_component.html.erb
@@ -1,0 +1,1 @@
+<%= render GovukComponent::SummaryListComponent.new(rows:, card: { title: }) %>
--- a/app/components/npq_summary_component.rb
+++ b/app/components/npq_summary_component.rb
@@ -1,0 +1,37 @@
+# frozen_string_literal: true
+
+class NpqSummaryComponent < ViewComponent::Base
+  include ActiveModel::Model
+
+  attr_accessor :qualification
+
+  def rows
+    [
+      {
+        key: {
+          text: "Awarded"
+        },
+        value: {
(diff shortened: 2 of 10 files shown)


spec/components/npq_summary_component_spec.rb
```diff
@@ -0,0 +1,15 @@
...
+RSpec.describe NpqSummaryComponent, type: :component do
+  pending "add some examples to (or delete) #{__FILE__}"
+
+  # it "renders something useful" do
+  #   expect(
+  #     render_inline(described_class.new(attr: "value")) { "Hello, components!" }.css("p").to_html
+  #   ).to include(
+  #     "Hello, components!"
+  #   )
+  # end
```
malcolmbaig: Worth either writing a simple spec or deleting this file I think. Appreciate that the component is quite simple and adequately tested via the system spec.

malcolmbaig (approved):
Bit of minor feedback, otherwise LG 👍🏼

spec/components/npq_summary_component_spec.rb
felixclack: Good point. I'll remove it.

PR #70, opened 2023-03-20
Add a common interface for qualifications

felixclack:
The DQT API returns a variety of qualification types and uses separate
attributes on the response to represent them.

The UI we are building treats displays all qualifications in the same
format.

I opted to have the API client transform the incoming data into a
consistent interface that makes the UI simpler and easier to work with.

### Link to Trello card

https://trello.com/c/TGRCucwG/822-order-the-qualifications-by-awarded-date

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

Screenshot 2023-03-20 at 12 15 01 pm


app/components/qualification_summary_component.rb
```diff
@@ -0,0 +1,97 @@
...
+
+class QualificationSummaryComponent < ViewComponent::Base
+  include ActiveModel::Model
+
+  attr_accessor :qualification
+
+  delegate :awarded_at, :details, :name, :type, to: :qualification
+
+  def rows
+    return itt_rows if name == "Initial teacher training (ITT)"
```
malcolmbaig: Can we check for type == :itt here?

app/components/qualification_summary_component.rb
```diff
@@ -0,0 +1,97 @@
...
+          text: details.qualification.name
+        }
+      },
+      { key: { text: "ITT provider" }, value: { text: details.provider.name } },
+      {
+        key: {
+          text: "Training type"
+        },
+        value: {
+          text: details.programmeType
```
malcolmbaig: I think ideally the case of attributes/keys coming out of the API client would be consistent. Can we run a transform_keys on hashes prior to them being passed in as `details`?

app/lib/qualifications_api/teacher.rb
```diff
@@ -3,69 +3,61 @@ class Teacher
...
-        .map do |npq|
-          Struct.new(:name, :certificate_url, :type, :awarded_at).new(
-            npq["type"]["name"],
-            npq["certificateUrl"],
-            npq["type"]["code"],
-            npq["awarded"]&.to_date
+        .each do |npq|
+          @qualifications << Qualification.new(
+            awarded_at: npq.awarded.to_date,
+            details: npq,
```
malcolmbaig: To clarify, where do we use npq details?

felixclack: Good question. We don't need it.

malcolmbaig (approved):
👌🏼

PR #71, opened 2023-03-20
Add route constraint for AYTQ service

malcolmbaig:
### Context
We're planning to build a Check teacher records service. We've decided to implement it as part of this repo.
Following a pattern established by DFE-Digital/publish-teacher-training, we can use route constraints to ensure routes for a given service are only accessible via requests to the appropriate hostname. This allows the two services to live at different URLs and use the same codebase without any conflicts.

In this commit:

- Introduce the route constraint for AYTQ
- Move AYTQ routes into their own file, and apply the constraint
- Assume for now that support / staff routes will be shared between the two services
- Add HOSTING_DOMAIN to .env.test so that system specs pass

### Guidance to review
The system specs passing is a pretty clear indicator that routes are accessible as expected. Manually altering the value of `HostingEnvironment.aytq_domain` will render them inaccessible.

### Link to Trello card
https://trello.com/c/c6MZPvFp

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased main
- [ ] Cleaned commit history

### DIFF
--- a/lib/route_constraints/access_your_teaching_qualifications_constraint.rb
+++ b/lib/route_constraints/access_your_teaching_qualifications_constraint.rb
@@ -1,0 +1,8 @@
+module RouteConstraints
+  class AccessYourTeachingQualificationsConstraint
+    def matches?(request)
+      request.host.in?(HostingEnvironment.aytq_domain) ||
+        request.host.include?("aytq-review-pr")
+    end
+  end
+end
--- a/.env.test
+++ b/.env.test
@@ -1,4 +1,5 @@
 GOVUK_NOTIFY_API_KEY=override-locally
+HOSTING_DOMAIN=http://127.0.0.1
 HOSTING_ENVIRONMENT=local
 IDENTITY_API_DOMAIN=override-locally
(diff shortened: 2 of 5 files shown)


lib/route_constraints/access_your_teaching_qualifications_constraint.rb
```diff
@@ -0,0 +1,8 @@
+module RouteConstraints
+  class AccessYourTeachingQualificationsConstraint
+    def matches?(request)
+      request.host.in?(HostingEnvironment.aytq_domain) ||
+        request.host.include?("aytq-review-pr")
```
felixclack, line 5:
```suggestion
      request.host.in?([HostingEnvironment.aytq_domain, "aytq-review-pr"])
```

felixclack (approved):

lib/route_constraints/access_your_teaching_qualifications_constraint.rb
malcolmbaig: The review app hostname has a dynamic component (eg—aytq-review-pr-71) hence the two different kinds of check.

PR #74, opened 2023-03-21
Fetch Induction data from the API

felixclack:
The DQT API now supplies data on the teacher's induction qualification.

This data differs from the other qualifications and due to the way the
GOVUK view components are built, we need to create a custom version of
the summary list component. Allowing us to accommodate the induction
history entries.

### Link to Trello card

https://trello.com/c/EiGFxMOL/824-display-induction-data-from-the-api

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

capybara-202303221009108915176521

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -10,6 +10,14 @@ inherit_mode:
   merge:
     - Exclude
 
+Style/BlockDelimiters:
+  Exclude:
+    - app/components/induction_summary_component.rb
+
+Style/MethodCalledOnDoEndBlock:
+  Exclude:
+    - app/components/induction_summary_component.rb
+
 Style/NumericLiterals:
   Exclude:
     - db/schema.rb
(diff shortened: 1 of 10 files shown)


.rubocop.yml
```diff
@@ -10,6 +10,14 @@ inherit_mode:
   merge:
     - Exclude
 
+Style/BlockDelimiters:
+  Exclude:
+    - app/components/induction_summary_component.rb
+
+Style/MethodCalledOnDoEndBlock:
+  Exclude:
+    - app/components/induction_summary_component.rb
```
malcolmbaig, line 19: If these aren't temporary, might be worth disabling the rules locally in the file itself.

felixclack: Sure, can do. What's the reason behind your preference here?

malcolmbaig: Actually nvm, can see the benefit of keeping stuff like this in a single place. Ignore! 😄

malcolmbaig (approved):

PR #79, opened 2023-03-22
Add route constraint for Check teacher records service

malcolmbaig:
### Context
This new service will be implemented as part of this repo, and made accessible via a different URL.

This PR creates a separate routing namespace for the service.

- Make HostingEnvironment aware of the Check teacher records host via the CHECK_RECORDS_DOMAIN env var
- Add a route constraint for the service which compares the host of the incoming request with the value of HostingEnvironment#check_records_domain
- Define a route set for Check teacher records in a separate file, as we do for the AYTQ routes
- Add a blank sign in page as a placeholder for the root path.

### Guidance to review
At this point, just needs a view on whether the approach is sensible and naming of things makes sense. There'll be actual  stuff to test once we get this preliminary work done.

### Link to Trello card
https://trello.com/c/REh4Q6Nk

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history


app/models/hosting_environment.rb
```diff
@@ -11,6 +11,10 @@ def self.aytq_domain
     host
   end
 
+  def self.check_records_domain
+    ENV.fetch("CHECK_RECORDS_DOMAIN", "localhost")
```
malcolmbaig, line 15: This is a temporary fallback until the prod-like environments are correctly setting CHECK_RECORDS_DOMAIN when being deployed.

felixclack: I've been wondering if we should start using the Rails credentials files more.

We'd have better control of configuration variables in all environments and be able to be record changes to them in our version control without leaking them.

What do you think?

felixclack (approved):

app/models/hosting_environment.rb
malcolmbaig: I think it's a good idea, definitely worth discussing. I'd like to not use Azure for this at all, but I think there'd be some resistance from the wider team to deviating from that model. Especially as production env var settings are locked down by Azure's access control.

PR #88, opened 2023-03-29
Add console1984 & audits1984

felixclack:
We have encrypted user PII now and want to protect access to it via the
Rails console.

There are 2 libraries that can help us with this. `console1984` provides
the ability to decrypt and encrypt the data.

`audits1984` provides auditing of these decryption actions along with an
interface to review the logs.

Something to note is that we aren't able to connect the auditing to our
authentication system.

The console asks for details about the person making the decrypt request
along with details of consent to view it. However, this is entirely
reliant on the user providing truthful details here.

This doesn't protect against bad actors.

### Link to Trello card

https://trello.com/c/goRspcE7/853-add-db-encryption

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

Example console session:

Screenshot 2023-03-29 at 11 08 48 am

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -42,6 +42,51 @@ ActiveRecord::Schema[7.0].define(version: 2023_03_29_090435) do
     t.index ["blob_id", "variation_digest"], name: "index_active_storage_variant_records_uniqueness", unique: true
   end
 
+  create_table "audits1984_audits", force: :cascade do |t|
+    t.integer "status", default: 0, null: false
+    t.text "notes"
+    t.bigint "session_id", null: false
+    t.bigint "auditor_id", null: false
+    t.datetime "created_at", null: false
+    t.datetime "updated_at", null: false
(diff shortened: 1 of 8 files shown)


db/schema.rb
```diff
@@ -42,6 +42,54 @@
...
+
+  create_table "console1984_users", force: :cascade do |t|
+    t.string "username", null: false
+    t.datetime "created_at", null: false
+    t.datetime "updated_at", null: false
+    t.index ["username"], name: "index_console1984_users_on_username"
+  end
+
+  create_table "data_migrations", primary_key: "version", id: :string, force: :cascade do |t|
+  end
```
malcolmbaig, line 91: Missed from https://github.com/DFE-Digital/access-your-teaching-qualifications/pull/87/? Worth putting this in a separate commit, given it's not related to the auditing stuff.

malcolmbaig (approved):

PR #109, opened 2023-04-13
Check for presence of mandatory qualifications data in API data

malcolmbaig:
There are scenarios where this may not be present (eg - test data on non-prod environments), so wrap the code which handles these with a conditional check on their presence.

### DIFF
--- a/app/lib/qualifications_api/teacher.rb
+++ b/app/lib/qualifications_api/teacher.rb
@@ -60,13 +60,15 @@ module QualificationsApi
         )
       end
 
-      @qualifications << api_data.mandatory_qualifications.map do |mq|
-        Qualification.new(
-          awarded_at: mq.awarded.to_date,
-          details: mq,
-          name: "Mandatory qualification (MQ): specialist teacher",
-          type: :mandatory
-        )
+      if api_data.mandatory_qualifications.present?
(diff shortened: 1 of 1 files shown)


app/lib/qualifications_api/teacher.rb
```diff
@@ -60,13 +60,15 @@ def qualifications
...
 
-      @qualifications << api_data.mandatory_qualifications.map do |mq|
-        Qualification.new(
-          awarded_at: mq.awarded.to_date,
-          details: mq,
-          name: "Mandatory qualification (MQ): specialist teacher",
-          type: :mandatory
-        )
+      if api_data.mandatory_qualifications.present?
+        @qualifications << api_data.mandatory_qualifications.map do |mq|
```
felixclack, line 64: Isn't this happening as a result of the changes to the API?

My PR should address this issue. Would be interested to see if the scenario you encountered still exists.

malcolmbaig: Yes, I think you're right 👍🏼

malcolmbaig:
Closing — fixed more comprehensively by https://github.com/DFE-Digital/access-your-teaching-qualifications/pull/108

PR #110, opened 2023-04-13
Update Identity auth to work without Devise

malcolmbaig:
### Context

Devise doesn't support the use of its [omniauthable module](https://github.com/heartcombo/devise/wiki/OmniAuth-with-multiple-models) in more than one model. The documentation doesn't elaborate on why, but one obvious side effect is the failure to create the appropriate `/auth/:provider` routes for subsequently added providers, thus preventing the initiation of the request phase. We want to add Dfe Sign In (DSI) to this codebase for use in the Check teacher record service, so we need a way to implement two OAuth-based auth mechanisms alongside each other without relying on Devise.

You can probably hack together a solution which follows the advice in the wiki link above and still makes use of certain Devise helpers/controllers. I've made the call here that the implementation is easier to follow if we rework it to rely solely on the Omniauth gem and directly setting/clearing the session for authenticated users.

The commits provide a comprehensive outline, but the summary is:

- Remove usage of any Devise-specific functionality
- Use the Omniauth gem to configure the Identity strategy
- Handle the contents of the session directly rather than using `sign_in` and `sign_out` methods.

### Guidance to review
- Are the changes clear?
- Are we happy to proceed with a similar implementation for DSI?

### Link to Trello card
https://trello.com/c/QIjrfSJm

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history


app/controllers/identity_users_controller.rb
```diff
@@ -1,4 +1,4 @@
-class IdentityUsersController < ApplicationController
+class IdentityUsersController < QualificationsInterfaceController
   before_action :authenticate_user!
```
felixclack, line 2: We could drop this line as it duplicates the before_action in the parent.

app/helpers/application_helper.rb
```diff
@@ -20,7 +20,7 @@ def navigation
         if current_staff
           header.with_navigation_item(href: main_app.staff_sign_out_path, text: "Sign out")
         end
-      else
+      when "qualifications"
```
felixclack, line 23: Do we have qualifications namespace yet?

felixclack (approved):
Overall it looks good. Just left a couple of minor questions.

app/helpers/application_helper.rb
malcolmbaig: This is a bit of a hack, admittedly, and takes advantage of the fact that most of the service's functionality is reached without navigating away from the hub page served at `/qualifications`. We should have a proper namespace for AYTQ. I'll card it up.

PR #123, opened 2023-04-24
Add a namespace for AYTQ

felixclack:
The app is responsible for serving both AYTQ and Check Records.

A namespace for Check Records already exists but we don't have a
corresponding one for AYTQ.

I opted to leave the existing sign in routes outside the namespace as
we're still working on an integration with Identity.

### Link to Trello card

https://trello.com/c/fvsM7qiT/8-properly-namespace-access-your-teaching-qualifications

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased main
- [ ] Cleaned commit history


config/routes/aytq.rb
```diff
@@ -6,17 +6,18 @@
 get "/sign-in", to: "users/sign_in#new"
 get "/sign-out", to: "users/sign_out#new"
```
malcolmbaig: We could move sign-in / sign-out into the namespace, and update the controller namespace for the callback. Shouldn't affect Identity integration?

malcolmbaig:
This might be a good opportunity to reorganise system specs as well, including support files (like the extracted common steps).

config/routes/aytq.rb
felixclack: Yeah, that's a good point. Updated.

config/routes.rb
```diff
@@ -31,4 +32,6 @@
 
   constraints(RouteConstraints::AccessYourTeachingQualificationsConstraint.new) { draw(:aytq) }
   constraints(RouteConstraints::CheckRecordsConstraint.new) { draw(:check_records) }
+
+  root to: redirect("/qualifications/sign-in")
```
malcolmbaig: This likely an issue if navigating to the root of the Check service. I wonder if defining different `root` in each of the `aytq` and `check_records` route files will do the job?

felixclack: If we don't define a root then the default Rails page will display.

felixclack: Also, it doesn't affect the namespace root URLs.

malcolmbaig: The above code is fine if you navigate to `https://quals.gov.uk/`, but I think it'll error if you go to `https://check-records.gov.uk/` (ie—the base url of the Check service). Because of the route constraints, it won't have access to `/qualifications/sign-in`. 

We're also not serving the two services from two different hostnames in dev, like we would in the prod-like environments. It highlights something we're missing — a setup similar to what Publish does [here](https://github.com/DFE-Digital/publish-teacher-training/blob/main/guides/setup-development.md#configuring-local-domains).

I think we should card up adding something like that to our dev workflow (and document it in the readme). In the meantime maybe it's better to live with the default Rails page for now?

malcolmbaig: (https://trello.com/c/ylTD364t/18-add-dev-setup-to-serve-up-check-and-aytq-from-different-hostnames)

.env.test
```diff
@@ -1,6 +1,6 @@
 GOVUK_NOTIFY_API_KEY=override-locally
-CHECK_RECORDS_DOMAIN=http://127.0.0.1
-HOSTING_DOMAIN=http://127.0.0.1
+CHECK_RECORDS_DOMAIN=http://check_records.lvh.me
+HOSTING_DOMAIN=http://qualifications.lvh.me
```
malcolmbaig: Nice idea, however this introduces a dependency on a 3rd party service which hasn't been [entirely reliable](https://stackoverflow.com/a/67877376) in its recent history.

I've done some quick testing and it looks like we can use (for example):

```suggestion
CHECK_RECORDS_DOMAIN=check-records.localhost
HOSTING_DOMAIN=qualifications.localhost
```

With the corresponding tweaks to config along the lines of what you've done elsewhere in this PR, this should give us passing tests and a workable solution for different hostnames in dev.

config/environments/test.rb
```diff
@@ -52,4 +52,7 @@
   config.action_mailer.delivery_method = :test
   config.action_mailer.default_options = { from: "mail@example.com" }
   config.action_mailer.default_url_options = { host: "localhost", port: 3000 }
+  config.hosts << "check_records.lvh.me"
+  config.hosts << "qualifications.lvh.me"
```
malcolmbaig: Do we have access to the env vars set in .env.test at this point?

config/environments/test.rb
```diff
@@ -52,4 +52,7 @@
...
+  config.hosts << "www.example.com"
```
malcolmbaig: Is this required?

terraform/workspace_variables/preprod.tfvars.json
```diff
@@ -1,11 +1,11 @@
...
+  "environment_name": "preprod",
+  "key_vault_name": "s165t01-aytq-pp-kv",
+  "resource_group_name": "s165t01-aytq-pp-rg",
+  "storage_account_name": "s165t01aytqtfstatepp",
+  "enable_blue_green": true,
+  "app_service_plan_sku": "S1",
+  "keyvault_logging_enabled": true,
+  "storage_log_categories": [],
+  "resource_prefix": "s165t01-aytq"
+}
```
malcolmbaig, line 11: Some stray terraform changes here

.env.test
felixclack: Fine by me.

config/environments/test.rb
felixclack: Yeah, this is for the request specs.

terraform/workspace_variables/preprod.tfvars.json
felixclack: They got merged without linting. All linting fails now.

config/environments/test.rb
felixclack: No, it doesn't work without hardcoding them for some reason.

.env.test
```diff
@@ -1,6 +1,6 @@
 GOVUK_NOTIFY_API_KEY=override-locally
-CHECK_RECORDS_DOMAIN=http://127.0.0.1
-HOSTING_DOMAIN=http://127.0.0.1
+CHECK_RECORDS_DOMAIN=http://check_records.localhost
+HOSTING_DOMAIN=http://qualifications.localhost
```
malcolmbaig, line 3: Minor, and can be addressed later, but if we drop the protocol here, we can use these env vars in the various configuration bits below (and simply interpolate a string where a protocol is needed, ie—the capybara app host).

app/models/hosting_environment.rb
```diff
@@ -4,15 +4,15 @@ module HostingEnvironment
   PRODUCTION_URL = I18n.t("service.url")
 
   def self.host
-    ENV.fetch("HOSTING_DOMAIN")
+    ENV.fetch("HOSTING_DOMAIN", "qualifications.lvh.me")
```
malcolmbaig: I think we can drop the default values in this file — an environment which doesn't have these set should probably fail early. WDYT?

felixclack: Personally I'm not a fan of ENV variables for things like this. It tends to be a blocker for people because it is hidden away.

Would love to move to using Rails encrypted credentials to deal with this stuff but it is low priority given we don't need to onboard anyone on the project for the forseeable.

config/application.rb
```diff
@@ -49,5 +49,9 @@ class Application < Rails::Application
       base_controller_class: "SupportInterface::SupportInterfaceController"
     }
     config.console1984 = { ask_for_username_if_empty: true }
+
+    [ENV.fetch("HOSTING_DOMAIN"), ENV.fetch("CHECK_RECORDS_DOMAIN")].each do |domain|
+      config.hosts << domain
+    end
```
malcolmbaig: Instead of this, we should probably just have the following entries in config/environments/dev, to complement the ones already in config/environments/test:

```
 config.hosts << "check_records.localhost"
 config.hosts << "qualifications.localhost"
```

malcolmbaig (commented):
Looking good 👍🏼 . One thing to bear in mind is that this breaks Identity auth locally. I'll add a card to resolve.

config/application.rb
felixclack: What if the ENV variables don't match?

How would you handle setting these in production?

felixclack:
> Looking good 👍🏼 . One thing to bear in mind is that this breaks Identity auth locally. I'll add a card to resolve.

Strange. It's working locally for me.

config/application.rb
malcolmbaig: > What if the ENV variables don't match?

Could you potentially do this instead:
```
  config.hosts << ENV.fetch("CHECK_RECORDS_DOMAIN")
  config.hosts << ENV.fetch("HOSTING_DOMAIN")
```
That appears to boot just fine with the correct values being set.

> How would you handle setting these in production?

It seems that this is empty in prod [by default](https://guides.rubyonrails.org/configuring.html#actiondispatch-hostauthorization), and setting it introduces a `Host` header check. If we opt to start doing that I think we should consider it in a separate PR, but it doesn't look like we have a motivation to do so currently.

config/environments/test.rb
malcolmbaig: It's the protocol in the env var value that's causing an issue (see https://github.com/DFE-Digital/access-your-teaching-qualifications/pull/123#discussion_r1179014302)

config/application.rb
felixclack: Cool. Happy to turn this feature off for now.

malcolmbaig (approved):
Worth updating the commit message to remove the reference to lvh.me 👍🏼

PR #130, opened 2023-04-26
Encrypt sensitive DSI User fields

felixclack:
Whenever we store sensitive data on an individual, we want to encrypt
it.

This change adds encryption to the DSI User record for PII. Currently,
the only affected fields are email, first name and last name.

The encrypted version of the fields are longer and don't reliably fit
within the default 255 character limit, so we increase the length of the
column to accommodate this.

### Link to Trello card

https://trello.com/c/SKDNWe4i/16-add-db-encryption-for-dsi-user-records

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/app/models/dsi_user.rb
+++ b/app/models/dsi_user.rb
@@ -1,4 +1,7 @@
 class DsiUser < ApplicationRecord
+  encrypts :email, deterministic: true
+  encrypts :first_name, :last_name
+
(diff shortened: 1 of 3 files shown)


malcolmbaig (approved):
Thanks! 👍🏼

PR #132, opened 2023-04-26
Add session expiry for DSI authentication

felixclack:
When authenticating in the check records service using the DSI method,
there is no token expiry.

A long-lived token like this has an increased security risk over a
short-lived one.

To resolve this, we are proposing adding an arbitrary time out on the
session, starting with a 2 hour limit.

There is nothing magic about the choice of 2 hours, so if it turns out
that you think there is a better window, then feel free to change.

The implementation sets a hard-limit of 2 hours from a successful
response from the identity auth provider. This means that if the user
has a single session longer than 2 hours, they will be required to sign
in again to continue accessing the service.

### Link to Trello card

https://trello.com/c/HbzmgxeL/12-add-dsi-session-expiry

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/check_records/check_records_controller.rb
+++ b/app/controllers/check_records/check_records_controller.rb
@@ -1,6 +1,7 @@
 module CheckRecords
   class CheckRecordsController < ApplicationController
     before_action :authenticate_dsi_user!
+    before_action :handle_expired_session!
 
     layout "check_records_layout"
 
@@ -18,6 +19,18 @@ module CheckRecords
 
     def dsi_user_signed_in?
       !!current_dsi_user
+    end
+
+    def handle_expired_session!
+      if session[:dsi_user_session_expiry].nil?
+        redirect_to check_records_sign_out_path
+        return
+      end
+
+      if Time.zone.at(session[:dsi_user_session_expiry]).past?
+        flash[:warning] = "Your session has expired. Please sign in again."
+        redirect_to check_records_sign_out_path
+      end
     end
   end
 end
--- a/app/controllers/check_records/omniauth_callbacks_controller.rb
+++ b/app/controllers/check_records/omniauth_callbacks_controller.rb
@@ -5,7 +5,8 @@ class CheckRecords::OmniauthCallbacksController < ApplicationController
 
   def dfe
     @dsi_user = DsiUser.create_or_update_from_dsi(request.env["omniauth.auth"])
-    @dsi_user.begin_session!(session)
+    session[:dsi_user_id] = @dsi_user.id
+    session[:dsi_user_session_expiry] = 2.hours.from_now.to_i
 
     redirect_to check_records_root_path
   end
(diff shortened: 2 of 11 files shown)


app/controllers/check_records/omniauth_callbacks_controller.rb
```diff
@@ -7,6 +7,7 @@ def dfe_bypass
     redirect_to check_records_root_path unless CheckRecords::DfESignIn.bypass?
     @dsi_user = DsiUser.create_or_update_from_dsi(request.env["omniauth.auth"])
     @dsi_user.begin_session!(session)
+    session[:dsi_user_token_expiry] = 2.hours.from_now.to_i
```
malcolmbaig: We could make this part of `#begin_session!`

spec/system/check_records/user_access_token_expires_spec.rb
```diff
@@ -0,0 +1,30 @@
+require "rails_helper"
+
+RSpec.feature "DSI session expiry", type: :system do
+  include CommonSteps
+  include CheckRecord::AuthenticationSteps
```
malcolmbaig:
```suggestion
  include CheckRecords::AuthenticationSteps
```

app/controllers/check_records/omniauth_callbacks_controller.rb
felixclack: I'm not sure I follow the logic for the begin_session method. It seems out of scope for a DB backed model to also handle the session. Surely that is the responsibility of the controller?

malcolmbaig: That's a hangover from another service that I based this implementation on. Not particularly attached to the approach, but I think any session handling code related to this callback should be co-located, whether it's here or in that method. Happy for the method to be removed and its logic added here if you prefer 👍🏼

app/controllers/check_records/check_records_controller.rb
```diff
@@ -1,6 +1,7 @@
 module CheckRecords
   class CheckRecordsController < ApplicationController
     before_action :authenticate_dsi_user!
+    before_action :handle_expired_token!
```
malcolmbaig: Just realised something about this PR — we're using the language of token expiry, mirroring the implementation for Identity. But in practice that isn't what's happening — it isn't that a token is no longer valid, we're just manually setting an explicit expiry on the session we create for DSI, so that the timeout isn't ambiguous.

I think we should reword this implementation along the lines of eg— `handle_expired_dsi_session`, or `check_dsi_session_expiry`.

spec/system/check_records/user_access_token_expires_spec.rb
```diff
@@ -0,0 +1,30 @@
+require "rails_helper"
+
+RSpec.feature "DSI session expiry", host: :check_records, type: :system do
+  include CommonSteps
+  include CheckRecords::AuthenticationSteps
+
+  after { travel_back }
+
+  scenario "Access token expires", test: :with_stubbed_auth do
```
malcolmbaig: "access token" needs renaming to "session" in this file

malcolmbaig (approved):
One bit of renaming left, otherwise LG 👍🏼

PR #144, opened 2023-05-04
Initial implementation of TRN search and results page

malcolmbaig:
### Context

The Check records service allows employers to look up teachers by their TRN. 

A basic implementation based on the current state of the prototype. This covers:
- The search box
- Controller actions for a new search and the results page
- A bare-bones results page, which can be built up over subsequent PRs

### Guidance to review

Sign in to the Check service locally to manually test. You'll need a valid QUALIFICATIONS_API_FIXED_TOKEN in your `.env.development.local` (DM me).

### Link to Trello card
https://trello.com/c/ylTD364t

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history


app/controllers/check_records/search_controller.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+module CheckRecords
+  class SearchController < CheckRecordsController
+    def new
+    end
+
+    def result
```
felixclack: Could this be the `show` action instead?

I generally avoid custom actions in Rails controllers in order to stick with RESTful conventions and make it easier for the future maintainer.

app/views/check_records/search/new.html.erb
```diff
@@ -0,0 +1,12 @@
...
+    <h1 class="govuk-heading-l">Find a teacher record</h1>
+
+    <%= form_with url: check_records_result_path, method: :get do |f| %>
+      <%= f.govuk_text_field :trn, class: "govuk-input--width-20", label: { text: "Teacher reference number (TRN)" } %>
+      <%= f.govuk_submit "Search" %>
+    <% end %>
+  </div>
+</div>
+
+
```
felixclack, line 12:
```suggestion
```

app/controllers/check_records/search_controller.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+module CheckRecords
+  class SearchController < CheckRecordsController
```
felixclack, line 4: I wonder if it would make more sense for this to be named `TeachersController`, as that is the resource that this returns. What do you think?

app/views/check_records/search/result.html.erb
```diff
@@ -0,0 +1,29 @@
+
```
felixclack:
```suggestion
```

app/views/check_records/search/result.html.erb
```diff
@@ -0,0 +1,29 @@
...
+    <h2 class="govuk-heading-m">Early years teacher status (EYTS)</h2>
+    <h2 class="govuk-heading-m">Induction</h2>
+    <h2 class="govuk-heading-m">Initial teacher training (ITT)</h2>
+    <h2 class="govuk-heading-m">Mandatory qualifications (MQ)</h2>
+    <h2 class="govuk-heading-m">National professional qualifications (NPQ)</h2>
+  </div>
+</div>
+
+
+
```
felixclack:
```suggestion
```

app/controllers/check_records/omniauth_callbacks_controller.rb
```diff
@@ -8,7 +8,7 @@ def dfe
     session[:dsi_user_id] = @dsi_user.id
     session[:dsi_user_session_expiry] = 2.hours.from_now.to_i
 
-    redirect_to check_records_root_path
+    redirect_to check_records_search_path
```
felixclack: The root path is now the search path. I don't think we need to change this line.

felixclack (approved):
LGTM. A few minor suggestions but they are just preferences. Nothing major.

app/controllers/check_records/omniauth_callbacks_controller.rb
malcolmbaig: Yes, good spot — did this prior to the path updates.

app/controllers/check_records/search_controller.rb
malcolmbaig: That would give us TeachersController#new which doesn't seem like the right fit for a page which prompts you to start a new search. The action rename suggested above leaves things in a pretty clear state I think.

PR #166, opened 2023-06-08
Add SSL checking to statuscake

nickdevdfe:
it appears that the domain K:V pair is missing from the production.tfvars.json


nickdevdfe (commented):
Setting a 'domain' in tfvars triggers conditional terraform resources, so added a new varaible just for Statuscake SSL checks.

saliceti (approved):

smneal (approved):

PR #167, opened 2023-06-08
Enable searching for teachers by last name and DOB

felixclack:
The designs modified the search function to use last name and date of
birth instead of TRN.

There is a new endpoint that we can use to search.

We can use the previous TRN search code to display the actual teacher
record with qualifications.

### Link to Trello card

https://trello.com/c/Liqqvd28/49-name-and-dob-search

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased main
- [ ] Cleaned commit history


app/controllers/check_records/search_controller.rb
```diff
@@ -6,17 +6,24 @@ def new
...
-
-      begin
-        client = QualificationsApi::Client.new(token: ENV["QUALIFICATIONS_API_FIXED_TOKEN"])
-        @teacher = client.teacher(trn: params[:trn])
-        @npqs = @teacher.qualifications.filter(&:npq?)
-        @other_qualifications =
-          @teacher.qualifications.filter { |qualification| !qualification.npq? }
-      rescue QualificationsApi::TeacherNotFoundError
-        render "not_found"
+      if params[:last_name].blank? && params[:date_of_birth].blank?
```
malcolmbaig: I think a form object would make sense here, so we can make the validations a bit more robust.

Currently, leaving either of these inputs blank results in two different exceptions.

app/assets/stylesheets/main.scss
```diff
@@ -51,3 +51,47 @@ $govuk-assets-path: "/";
 .app__two-thirds {
   width: 66.66%;
 }
+
+.app-list {
```
malcolmbaig: Is there a more descriptive name we can give this class?

malcolmbaig (changes requested):

app/assets/stylesheets/main.scss
felixclack: I grabbed these from the prototype.

Happy to give renaming them a go.

app/assets/stylesheets/main.scss
```diff
@@ -51,3 +51,47 @@ $govuk-assets-path: "/";
 .app__two-thirds {
   width: 66.66%;
 }
+
+.search-results {
```
malcolmbaig, line 55: Do you think it's worth namespacing Check-specific styles at this stage?

felixclack: A better approach would probably be to serve different CSS/JS for each service, to avoid people downloading styles and JS that will never be used.

I can do that as a follow up.

app/models/search.rb
```diff
@@ -0,0 +1,7 @@
+class Search
+  include ActiveModel::Model
+
+  attr_accessor :date_of_birth, :last_name
+
+  validates :date_of_birth, :last_name, presence: true
```
malcolmbaig: This hasn't quite done the trick — 

- An instance of search isn't being passed to the view, so no validation errors appear
- If any of the date fields are blank, you get a parse error. We'll need to validate the presence of these fields prior to parsing a date out of them.

app/models/search.rb
```diff
@@ -0,0 +1,108 @@
...
+
+  def date_of_birth=(date_fields)
+    return if date_fields.nil?
+
+    date_fields.map! { |f| f.is_a?(String) ? word_to_number(f.strip).to_s : f }
+
+    month_is_a_word = date_fields[1]&.to_i&.zero? && date_fields[1].length.positive?
+    date_fields[1] = word_to_month_number(date_fields[1]) if month_is_a_word
+
+    @date_of_birth = Struct.new(:year, :month, :day).new(*date_fields)
```
malcolmbaig: Search is currently broken because we pass the `to_s` of this struct to the API client. Any reason this can't be a `Date`?

Needs a manual test to confirm fix, system spec didn't pick this up. 

I think it's worth adding to the spec for this method an expectation that checks the result of calling `to_s` on the resulting `date_of_birth`.

felixclack: A Date object breaks the form. I’ll figure out the API.

felixclack: Done.

malcolmbaig (approved):

PR #222, opened 2023-07-03
Implement RP-initiated logout for OAuth provider

felixclack:
The Identity service implements the OpenIDConnect spec, which allows for
an RP-Initiated logout.

We aren't using the OpenIDConnect OmniAuth strategy currently, so don't
have access to the logout feature.

I considered re-implementing the strategy to use OpenIDConnect but
decided that it seemed overkill for what we need here, which is to
support a redirect on sign out.

Instead, I implemented the `other_phase` method that all OmniAuth
strategies call for non-matching requests.

We can match on the logout URL and trigger the correct redirect.

### Link to Trello card

https://trello.com/c/1unAP0QQ/1024-access-quals-id-snags

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased main
- [ ] Cleaned commit history


malcolmbaig:
Testing this locally, I get this error:

`error_description:The specified 'post_logout_redirect_uri' is invalid.`

Is there some additional configuration required on preprod Identity maybe?

felixclack:
> Testing this locally, I get this error:
> 
> `error_description:The specified 'post_logout_redirect_uri' is invalid.`
> 
> Is there some additional configuration required on preprod Identity maybe?

Make sure your local version of HOSTING_DOMAIN = `http://localhost:3000`

malcolmbaig:
Nice, works well 👍🏼 . Worth carding up a refactor to switch to an actual OIDC-based implementation you think? I think there's a fair bit of overlap with the existing strategy that inherits from `OmniAuth::Strategies::OAuth2`.

malcolmbaig (approved):

felixclack:
> Nice, works well 👍🏼 . Worth carding up a refactor to switch to an actual OIDC-based implementation you think? I think there's a fair bit of overlap with the existing strategy that inherits from `OmniAuth::Strategies::OAuth2`.

It's probably worthwhile just to set expectations for a future maintainer that Identity is an OIDC server.

PR #226, opened 2023-07-04
Update how qualifications are sorted

felixclack:
There are scenarios where a qualification has no associated date.

Currently, this causes the app to throw an exception as we sort
qualifications by date.

I landed on using a default value for the date in the sort function.
This allows us to position the qualifications without a date at the end
of the list, so it makes sense visually.

We don't want this date to be visible though as it is used purely for
the sort order, so I didn't set the default on the qualification itself.

Also, when the QTS and ITT qualifications share the same awarded at date,
the QTS should take priority.

### Link to Trello card

https://trello.com/c/1unAP0QQ/1024-access-quals-id-snags

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased main
- [ ] Cleaned commit history

### DIFF
--- a/spec/lib/qualifications_api/teacher_spec.rb
+++ b/spec/lib/qualifications_api/teacher_spec.rb
@@ -76,5 +76,53 @@ RSpec.describe QualificationsApi::Teacher, type: :model do
     it "sorts the qualifications in reverse chronological order by date of award" do
       expect(qualifications.map(&:type)).to eq(%i[NPQSL NPQML eyts qts induction mandatory itt])
     end
+
+    context "when a qualification has no awarded date" do
+      let(:api_data) do
+        {
+          "initial_teacher_training" => [
+            {
+              "qualification" => {
+                "name" => "BA"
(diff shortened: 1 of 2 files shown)


spec/lib/qualifications_api/teacher_spec.rb
```diff
@@ -76,5 +76,86 @@
...
+          ]
+        }
+      end
+
+      it "sorts the qualifications in reverse chronological order by date of award" do
+        expect(qualifications.map(&:type)).to eq(%i[itt NPQML])
+      end
+    end
+
+    context "when there are no MQs returned" do
```
malcolmbaig, line 120:
```suggestion
    context "when there are no qualifications returned" do
```
?

felixclack: This is specifically for no mandatory qualifications as that scenario causes an exception.

malcolmbaig (approved):
👍🏼

PR #234, opened 2023-07-06
Add OIDC strategy for Identity

felixclack:
The Identity service is an OpenID Connect provider. Changing the
strategy we use with OmniAuth can make this more obvious.

We currently have implemented the OAuth2 strategy for Identity and
extended it to take advantage of the OIDC logout feature.

Switching to an OIDC strategy makes it more obvious what features the
Identity service provides. It also, hopefully, makes things clearer for
any future maintainer of the code.

### Link to Trello card

https://trello.com/c/85JumIaw/1090-switch-quals-to-use-an-oidc-strategy-for-identity

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history


spec/support/system/qualifications_authentication_steps.rb
```diff
@@ -24,7 +24,7 @@ def given_identity_auth_is_mocked
         },
         credentials: {
           token: "token",
-          expires_at: (Time.zone.now + 1.hour).to_i
+          expires_in: 1.hour.to_i
```
malcolmbaig, line 25: Is this a recent API change?

config/initializers/omniauth.rb
```diff
@@ -36,10 +35,23 @@
...
+             scheme: "https",
+             secret: ENV.fetch("IDENTITY_CLIENT_SECRET")
+           },
+           discovery: true,
+           issuer: ENV.fetch("IDENTITY_API_DOMAIN"),
+           path_prefix: "/qualifications/users/auth",
+           pkce: true,
+           post_logout_redirect_uri: "#{ENV.fetch("HOSTING_DOMAIN")}/qualifications/sign-out",
+           response_type: :code,
+           scope: %w[email openid profile dqt:read]
```
malcolmbaig, line 56: Does making this switch allow us to simplify the existing logout implementation at all?

spec/support/system/qualifications_authentication_steps.rb
felixclack: The name of the attribute? Not sure when it changed.

lib/omniauth/strategies/identity.rb
malcolmbaig, line 1: By losing the separate strategy file, we also remove the customised `info` hash. In turn, `User.from_identity` doesn't retrieve and persist all data correctly.

I think the simplest thing to do here is to update the `.from_identity` method so that it retrieves date of birth, trn, and first/last name correctly from the auth data. It means the method needs direct knowledge about an externally-defined data structure, but as long as that doesn't leak out anywhere else in the codebase I think it's reasonable.

config/initializers/omniauth.rb
```diff
@@ -36,10 +35,23 @@
   Rails.application.config.middleware.use OmniAuth::Strategies::OpenIDConnect, options
 end
 
-# Identity setup
 Rails.application.config.middleware.use OmniAuth::Builder do
-  provider :identity,
-           ENV.fetch("IDENTITY_CLIENT_ID"),
-           ENV.fetch("IDENTITY_CLIENT_SECRET"),
-           { path_prefix: "/qualifications/users/auth" }
+  provider :openid_connect,
```
malcolmbaig, line 38: For consistency, I think we can align the DSI/Identity implementations a little more. Specifically by:

- Using `Rails.application.config.middleware.use OmniAuth::Strategies::OpenIDConnect, options` for both
- Defining two distinct options list for both and passing them in, or rewriting the DSI config above so that its options are inline (like you've done for Identity).

malcolmbaig (commented):
Nice one, thanks for picking up 👍🏼 . A few comments left inline.

lib/omniauth/strategies/identity.rb
felixclack: Yeah, I missed that. Will fix it up.

malcolmbaig: One thing I forgot to add: the `given_identity_auth_is_mocked` step in our system specs will need updating as well to reflect the slightly different payload we get back from DSI.

app/controllers/qualifications/users/omniauth_callbacks_controller.rb
```diff
@@ -8,7 +8,7 @@ def identity
         @user = User.from_identity(auth)
         session[:identity_user_id] = @user.id
         session[:identity_user_token] = auth.credentials.token
-        session[:identity_user_token_expiry] = auth.credentials.expires_at
+        session[:identity_user_token_expiry] = auth.credentials.expires_in.seconds.from_now.to_i
```
malcolmbaig, line 11: Testing this locally (using preprod Identity), I still get expires_at back, leading to an exception at this line. Any idea why?

felixclack:
This is strange because I get the opposite error on the main branch. 

On July 11, 2023, "codeclimate[bot]" ***@***.***> wrote:
> @malcolmbaig commented on this pull request.
>

> In
> app/controllers/qualifications/users/omniauth_callbacks_controller.rb
> <https://github.com/DFE-Digital/access-your-teaching-
> qualifications/pull/234#discussion_r1259662861>:
>
> > @@ -8,7 +8,7 @@ def identity @user = User.from_identity(auth)
> session[:identity_user_id] = @user.id session[:identity_user_token] =
> auth.credentials.token - session[:identity_user_token_expiry] =
> auth.credentials.expires_at + session[:identity_user_token_expiry] =
> auth.credentials.expires_in.seconds.from_now.to_i
> Testing this locally (using preprod Identity), I still get expires_at
> back, leading to an exception at this line. Any idea why?
>

—
Reply to this email directly, view it on GitHub <https://github.com/DFE-
Digital/access-your-teaching-qualifications/pull/234#pullrequestreview-
1524153723>, or unsubscribe
<https://github.com/notifications/unsubscribe-
auth/AAAAYNRV74OZGSQVDGS7XU3XPVBIFANCNFSM6AAAAAA2AC2U7I>.
You are receiving this because you authored the thread.Message ID: <DFE-
Digital/access-your-teaching-
***@***.***>

config/initializers/omniauth.rb
```diff
@@ -35,13 +11,44 @@
...
+               identifier: ENV.fetch("DFE_SIGN_IN_CLIENT_ID", "example"),
+               port: dfe_sign_in_issuer_uri&.port,
+               redirect_uri: ENV.fetch("DFE_SIGN_IN_REDIRECT_URL", "example"),
+               scheme: dfe_sign_in_issuer_uri&.scheme,
+               secret: ENV.fetch("DFE_SIGN_IN_SECRET", "example")
+             },
+             discovery: true,
+             path_prefix: "/check-records/auth",
+             response_type: :code,
+             scope: %i[email organisation profile]
```
malcolmbaig, line 33: Missing the `issuer` option specified in the original list — needed to ensure the correct endpoint is used during discovery.

malcolmbaig:
> This is strange because I get the opposite error on the main branch. 


Sorry, ignore! Removed my original comment because I realised I hadn't tested it properly. Needed a server restart. All good as far as this attribute is concerned.

config/initializers/omniauth.rb
```diff
@@ -35,13 +12,47 @@
...
+             client_options: {
+               host: dfe_sign_in_issuer_uri&.host,
+               identifier: ENV.fetch("DFE_SIGN_IN_CLIENT_ID", "example"),
+               port: dfe_sign_in_issuer_uri&.port,
+               redirect_uri: ENV.fetch("DFE_SIGN_IN_REDIRECT_URL", "example"),
+               scheme: dfe_sign_in_issuer_uri&.scheme,
+               secret: ENV.fetch("DFE_SIGN_IN_SECRET", "example")
+             },
+             discovery: true,
+             issuer: ENV.fetch("DFE_SIGN_IN_ISSUER", "example"),
```
malcolmbaig: This has quite a specific expected format — need the port number appended like in the original code

felixclack: Strangely, just the URI works for Identity. I updated it so the value doesn't change from what is currently deployed.

malcolmbaig: Yep, Identity and DSI have different configurations, can view the issuer value for both at (for example):

https://dev-oidc.signin.education.gov.uk/.well-known/openid-configuration
https://preprod.teaching-identity.education.gov.uk/.well-known/openid-configuration

config/initializers/omniauth.rb
```diff
@@ -35,13 +12,46 @@
...
-           ENV.fetch("IDENTITY_CLIENT_SECRET"),
-           { path_prefix: "/qualifications/users/auth" }
+  provider :openid_connect,
+           name: :identity,
+           callback_path: "/qualifications/users/auth/identity/callback",
+           client_options: {
+             host: URI(ENV.fetch("IDENTITY_API_DOMAIN")).host,
+             identifier: ENV.fetch("IDENTITY_CLIENT_ID"),
+             port: 443,
+             redirect_uri: ENV.fetch("IDENTITY_REDIRECT_URI"),
```
malcolmbaig: We'll need this set on dev/test/preprod/prod so the app doesn't fall over.

config/initializers/omniauth.rb
```diff
@@ -35,13 +12,46 @@
...
-  Rails.application.config.middleware.use OmniAuth::Strategies::OpenIDConnect, options
+  dfe_sign_in_issuer_uri = URI(ENV.fetch("DFE_SIGN_IN_ISSUER", "example"))
+
+  Rails.application.config.middleware.use OmniAuth::Builder do
+    provider :openid_connect,
+             name: :dfe,
+             callback_path: "/check-records/auth/dfe/callback",
+             client_options: {
+               host: dfe_sign_in_issuer_uri&.host,
+               identifier: ENV.fetch("DFE_SIGN_IN_CLIENT_ID", "example"),
```
malcolmbaig: I wonder if it'd be better to set a precedent and drop the default "example' value in these options? Useful for the app to crash early (during an attempt at starting the server in the dev environment, for example) so that we can pick up any misconfiguration before getting to prod.

felixclack: Made the suggested change as separate PR as it isn't directly related to this change.

https://github.com/DFE-Digital/access-your-teaching-qualifications/pull/244

PR #273, opened 2023-07-27
Check organisation on sign in

felixclack:
When a DSI user signs in to the service, we want to ensure that they are
part of a known organisation before letting them access the service.

To start with, we decided to simply compare the organisation associated
with the authenticated user with a known organisation in our database.

The company registration number is the proposed preferred field to match
on.

Testing this requires being added to the test organisation in
development or pre-prod environments of Identity. You can request this
from the DSI team.

Assumptions:
- the basic interface for creating organisations in support is for testing purposes
- we will implement a way to seed production data in a future change

### Link to Trello card

https://trello.com/c/I2Wjvki4/71-update-dfe-signin-integration-to-check-against-agency-list

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

Screenshot 2023-08-23 at 8 43 47 am
Screenshot 2023-08-23 at 8 41 25 am


malcolmbaig:
Commit message and PR description reference the Identity team — should be DSI team.

app/controllers/check_records/omniauth_callbacks_controller.rb
```diff
@@ -5,8 +5,11 @@ class CheckRecords::OmniauthCallbacksController < ApplicationController
 
   def dfe
     @dsi_user = DsiUser.create_or_update_from_dsi(request.env["omniauth.auth"])
-    session[:dsi_user_id] = @dsi_user.id
-    session[:dsi_user_session_expiry] = 2.hours.from_now.to_i
+
+    if @dsi_user.valid_organisation?
+      session[:dsi_user_id] = @dsi_user.id
+      session[:dsi_user_session_expiry] = 2.hours.from_now.to_i
+    end
```
malcolmbaig, line 16: Should we set a flash message if they don't pass this check?

db/production/agency_list.csv
```diff
@@ -0,0 +1,340 @@
+Name,legacy_ID,include_as_agency?,Assumed Company Registration Number,Note(s)
```
malcolmbaig: For clarity, it may be worth removing the unused columns (`legacy_id, `include_as_agency?` and `Note(s)`) in a spreadsheet program.

felixclack: The reason I kept them was to make it easy to update this file in the future. If we start removing columns then it means modifying the spreadsheet, which introduces extra manual steps.

My preference would be to be able to drag and drop the new file here without having to open up a spreadsheet and create a custom export.

🤷

malcolmbaig: Makes sense 👍🏼

malcolmbaig (commented):
LG. Are we merging the company reg number stuff? Does it make sense to rework this to guid instead at this stage?

felixclack:
> LG. Are we merging the company reg number stuff? Does it make sense to rework this to guid instead at this stage?

That will require DSI coming up with a guid for organisations. It's not something we can do currently unless I'm missing something?

felixclack:
Currently paused while we figure out the best way forward with sign in.

PR #277, opened 2023-08-01
Use the provided certificate URL

felixclack:
Currently we construct the URL for a certificate by using the
certificate type.

There are reports from users of 404 pages and the exceptions point to a
timeout when fetching a certificate.

The working theory is that we're constructing a URL that doesn't exist
and so the API call hits the timeout.

There's no reason to construct the certificate URL in this fashion as
the API returns the path if the certificate is present.

We can rely on this provided value instead.

### DIFF
--- a/app/components/induction_summary_component.rb
+++ b/app/components/induction_summary_component.rb
@@ -5,7 +5,7 @@ class InductionSummaryComponent < ViewComponent::Base
 
   attr_accessor :qualification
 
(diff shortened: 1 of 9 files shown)


felixclack:
Merging this because we need a fix in production and can't wait any longer. However, please still review this in retrospect and I'll make changes in another PR if there's anything you see.

PR #286, opened 2023-08-08
Add smoke tests for AYTQ and Check

malcolmbaig:
### Context
We don’t currently have smoke tests for these services. We should add some using the ones in Refer serious misconduct as a reference.

- First two commits are tangential changes to make it easier to run the smoke spec locally when editing it
- Add smoke test spec file
- Add terraform/github action changes to run the smoke test after deploys 

### Guidance to review
A once-over of the changes to double-check if I've missed anything is probably sufficient. Will be testing this "live" once merged and iterating on any failures. Failures will (I think, with the current setup) stop prod deploys. This doesn't matter for Check, but if we need anything to go out for AYTQ urgently, we can deploy manually.

### Link to Trello card
https://trello.com/c/PMm6UKoG/87-add-smoke-tests-for-aytq-check

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased main
- [ ] Cleaned commit history

### DIFF
--- a/.env.development
+++ b/.env.development
@@ -1,11 +1,11 @@
 BYPASS_DSI=true
-CHECK_RECORDS_DOMAIN=check.localhost
+CHECK_RECORDS_DOMAIN=http://check.localhost:3000
 DFE_SIGN_IN_CLIENT_ID=checkrecordteacher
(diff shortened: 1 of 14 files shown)


felixclack (approved):
LGTM

PR #290, opened 2023-08-10
Support the teachers search returning a 500 response

felixclack:
There is a scenario where when searching for teacher records the API can
return a 500 response.

We don't currently handle this and present an unhelpful error page.

I opted to treat this as a no match scenario for the user, as it is
effectively the same thing to them.

### Link to Trello card

https://trello.com/c/XNwY3qMX/137-fix-missing-search-results

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased main
- [ ] Cleaned commit history

### DIFF
--- a/app/lib/qualifications_api/client.rb
+++ b/app/lib/qualifications_api/client.rb
@@ -72,11 +72,13 @@ module QualificationsApi
       raise(QualificationApi::InvalidTokenError) if response.status == 401
 
       results =
-        response.body["results"].map do |teacher|
+        response.body["results"]&.map do |teacher|
           QualificationsApi::Teacher.new(teacher)
         end
 
-      [response.body["total"], results]
+      results ||= []
+      total = response.body["total"] || 0
+      [total, results]
     end
 
     def client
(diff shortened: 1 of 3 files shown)


app/lib/qualifications_api/client.rb
```diff
@@ -72,11 +72,13 @@ def teachers(date_of_birth:, last_name:)
       raise(QualificationApi::InvalidTokenError) if response.status == 401
 
       results =
-        response.body["results"].map do |teacher|
+        response.body["results"]&.map do |teacher|
```
malcolmbaig, line 75: Does this warrant a code comment, given that it's not immediately obvious why the response body wouldn't contain a results field?

malcolmbaig:
>I opted to treat this as a no match scenario for the user, as it is
effectively the same thing to them.

Not sure about this — if it's a transient error, maybe it would be preferable to let them know in some way that the search failed and they should try again? Also, should we send something to Sentry when this happens?

WDYT @gpeng ?

felixclack:
Closing this as James said API error has been addressed.

PR #297, opened 2023-08-10
Reenable smoke tests on preprod

malcolmbaig:
DNS changes have been made so that Check is available at https://preprod.check-the-record-of-a-teacher.education.gov.uk 

We can now run smoke tests against this environment.


felixclack (approved):
:shipit:

PR #298, opened 2023-08-10
Update SKU

gpeng:
Deploys are failing because 'a feature' isn't supported. We've added a worker to the failing environments so likely that. Trying this to bring it into parity with preprod.


AbigailMcP (approved):
🤞

PR #310, opened 2023-08-16
Add default_url_options[:host] for mail

gpeng:
### Context

Set the default host

Notify emails are failing due to this not being set.

### Guidance to review

This relates to -> https://dfe-teacher-services.sentry.io/issues/4399021638/?project=4505227155996672&query=is%3Aunresolved&referrer=issue-stream&stream_index=2

We're passing the domain into the devise mail and it's failing when we try to send invite mails.

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -80,4 +80,5 @@ Rails.application.configure do
   config.active_record.dump_schema_after_migration = false
 
   config.action_mailer.delivery_method = :notify
+  config.action_mailer.default_url_options = { host: HostingEnvironment.aytq_domain }
 end


config/environments/production.rb
```diff
@@ -80,4 +80,5 @@
   config.active_record.dump_schema_after_migration = false
 
   config.action_mailer.delivery_method = :notify
+  config.action_mailer.default_url_options = { host: HostingEnvironment.aytq_domain }
```
malcolmbaig, line 83: Assuming we only ever access the support/staff functionality via AYTQ (and not Check), this should be fine. Will need revisiting if we ever start sending emails from Check.

malcolmbaig (approved):

felixclack (approved):
Approving on the assumption that we never send emails from Check. I believe we only send out staff invites for the support interface.

config/environments/production.rb
gpeng: Yeah. I wondered about that. At the moment we're only sending invites for staff users I think so it'll be fine for now.

gpeng:
Reverting for now as this is stopping `main` building.

PR #322, opened 2023-08-23
Display sanction details for a specific teacher

felixclack:
When a teacher has sanctions in the API response, we want to display
them as per the designs.

There is on-going work to update the wording used currently to describe each
sanction. See https://educationgovuk.sharepoint.com/:x:/s/TeacherServices/EZsyllhdkb9Giji9pHvGG1UBKseSEplO-KUUH-vseaRM7w
for details.

We don't have a full description or date attached to any of the
sanctions at the moment, so this is where the implementation diverges
from the designs.

Assumptions:
- the sanctions list won't change much so storing them as key-values in
  code will be the most efficient way of representing them
- we won't be getting any more details from the API about the sanctions
- we can ship a version of the sanctions without the descriptions and
  revisit this once they have been written

### Link to Trello card

https://trello.com/c/w4Xv9m5m/104-add-sanctions-to-the-show-page

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

Screenshot 2023-08-23 at 1 16 21 pm

### DIFF
--- a/app/models/sanction.rb
+++ b/app/models/sanction.rb
@@ -1,0 +1,43 @@
+class Sanction
+  include ActiveModel::Model
+
+  attr_accessor :type
+
+  SANCTIONS = {
+    "A13" => { title: "Suspension order - with conditions" },
+    "A14" => { title: "Suspension order - with conditions" },
+    "A18" => { title: "Conditional registration order - conviction of a relevant offence" },
+    "A19" => { title: "Suspension order" },
+    "A1A" => { title: "Prohibition by the General Teaching Council England (GTCE)" },
+    "A1B" => { title: "Prohibition by the General Teaching Council England (GTCE)" },
+    "A2" => { title: "Suspension order" },
+    "A20" => { title: "Suspension order - with conditions" },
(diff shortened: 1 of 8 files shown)


app/models/sanction.rb
```diff
@@ -0,0 +1,43 @@
...
+    "T3" => { title: "Prohibition by the Secretary of State - deregistered by GTC Scotland" },
+    "T4" => { title: "Prohibition by the Secretary of State  - refer to the Education Workforce Council, Wales" },
+    "T5" => { title: "Prohibition by the Secretary of State - refer to GTC Northern Ireland" },
+    "T6" => { title: "Secretary of State decision - no prohibition" },
+    "T7" => { title: "Section 128 barring direction" }
+  }.freeze
+
+  def title  
+    SANCTIONS[type][:title] if SANCTIONS[type]
+  end
```
malcolmbaig, line 42: Suggestion: worth adding a couple of simple unit tests for this.

malcolmbaig (approved):
Nice 👍🏼 , one minor comment left inline

PR #349, opened 2023-09-12
Fix Staff email invite

felixclack:
There is an issue when inviting a staff user where the check for the
AnonymousSupportUser class doesn't work as expected.

Rather than check on class, we can use duck-typing and check for the
presence of an expected method.

This means the code is less coupled to a specific class and is more
explicit about the expectations for the code branch.

### Link to Trello card

https://trello.com/c/BgI0CKVW/180-staff-user-email-invites-not-working

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased main
- [ ] Cleaned commit history

### DIFF
--- a/spec/system/support/staff_invite_as_staff_member_spec.rb
+++ b/spec/system/support/staff_invite_as_staff_member_spec.rb
@@ -1,0 +1,75 @@
+require 'rails_helper'
+
+RSpec.feature 'Staff support', type: :system do
+  include CommonSteps
+
+  scenario "Staff user invites another Staff member" do
+    given_the_service_is_open
+    and_a_staff_member_exists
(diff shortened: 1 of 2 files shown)


spec/system/support/staff_invite_as_staff_member_spec.rb
```diff
@@ -0,0 +1,75 @@
...
+
+    expect(message.subject).to eq("Invitation instructions")
+    expect(message.to).to include("invite@example.com")
+  end
+
+  def then_i_see_the_invited_staff_user
+    expect(page).to have_content("invite@example.com")
+  end
+end
+
```
gpeng, line 75: newline

gpeng (approved):

PR #365, opened 2023-09-20
Configure sign out of DSI via Omniauth strategy

steventux:
### Context

Currently our ‘Sign out' link only ends the users session in our service.
We also need to sign out of DfE Signin too otherwise the user is unable to easily sign back in as a different organisation.

Adds an OmniAuth strategy which includes the `id_token_hint` param in the `end_session_redirect_uri`. 
Necessary for the [node-oidc-provider](https://github.com/panva/node-oidc-provider/blob/main/lib/actions/end_session.js) library to end the session and redirect successfully.
    
The `id_token_hint` value is the `auth.credentials.id_token` returned on successful sign in. 
This value is stored in the user session and passed as a param in the DfE signout path.
    
Configures omniauth to use this strategy.
    
Adding the redirect route `/check-records/auth/dfe/sign-out` is also necessary to make the redirect work as the underlying `omniauth_openid_connect` gem [pattern matches on the `request_path`](https://github.com/omniauth/omniauth_openid_connect/blob/master/lib/omniauth/strategies/openid_connect.rb#L437) as part of the redirect validation.

### Link to Trello card

https://trello.com/c/D916iWr3/195-sign-out-via-dfe-signin

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/config/initializers/omniauth.rb
+++ b/config/initializers/omniauth.rb
@@ -1,5 +1,7 @@
 require "check_records/dfe_sign_in"
+require "omniauth/strategies/dfe_openid_connect"
 
+OmniAuth.config.add_camelization('dfe_openid_connect', 'DfEOpenIDConnect')
 OmniAuth.config.logger = Rails.logger
 OmniAuth.config.on_failure =
   proc { |env| AuthFailuresController.action(:failure).call(env) }
@@ -15,9 +17,11 @@ else
   dfe_sign_in_issuer_uri = URI(ENV.fetch("DFE_SIGN_IN_ISSUER", "example"))
 
   Rails.application.config.middleware.use OmniAuth::Builder do
-    provider :openid_connect,
+    provider :dfe_openid_connect,
              name: :dfe,
              callback_path: "/check-records/auth/dfe/callback",
+             logout_path: "/sign-out",
+             post_logout_redirect_uri: "#{ENV['CHECK_RECORDS_DOMAIN']}/check-records/sign-out",
              client_options: {
                host: dfe_sign_in_issuer_uri&.host,
                identifier: ENV["DFE_SIGN_IN_CLIENT_ID"],
--- a/lib/omniauth/strategies/dfe_openid_connect.rb
+++ b/lib/omniauth/strategies/dfe_openid_connect.rb
@@ -1,0 +1,27 @@
+# This strategy ensures that the id_token_hint param is included in the post_logout_redirect_uri.
+# The node-oidc-provider library requires this param to be present in order for the redirect to work.
+# See: https://github.com/panva/node-oidc-provider/blob/03c9bc513860e68ee7be84f99bfc9dc930b224e8/lib/actions/end_session.js#L27
+# See: https://github.com/omniauth/omniauth_openid_connect/blob/34370d655d39fe7980f89f55715888e0ebd7270e/lib/omniauth/strategies/openid_connect.rb#L423
+#
+module OmniAuth
+  module Strategies
+    class DfEOpenIDConnect < OmniAuth::Strategies::OpenIDConnect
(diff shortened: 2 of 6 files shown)


steventux:
[Sign out via DfE Signin](https://trello.com/c/D916iWr3/195-sign-out-via-dfe-signin)

gpeng (approved):
LGTM. Might be worth getting a second opinion from @malcolmbaig and/or @felixclack

lib/omniauth/strategies/dfe_openid_connect.rb
```diff
@@ -0,0 +1,27 @@
+# This strategy ensures that the id_token_hint param is included in the post_logout_redirect_uri.
+# The node-oidc-provider library requires this param to be present in order for the redirect to work.
+# See: https://github.com/panva/node-oidc-provider/blob/main/lib/actions/end_session.js#L27
+# See: https://github.com/omniauth/omniauth_openid_connect/blob/master/lib/omniauth/strategies/openid_connect.rb#L423
```
malcolmbaig: Nit: would it be worth replacing main/master with the SHAs of current HEAD in both cases, so the comment remains accurate as time goes on?

config/initializers/omniauth.rb
```diff
@@ -15,9 +17,11 @@
   dfe_sign_in_issuer_uri = URI(ENV.fetch("DFE_SIGN_IN_ISSUER", "example"))
 
   Rails.application.config.middleware.use OmniAuth::Builder do
-    provider :openid_connect,
+    provider :dfe_openid_connect,
              name: :dfe,
              callback_path: "/check-records/auth/dfe/callback",
+             logout_path: "/sign-out",
+             post_logout_redirect_uri: "#{ENV['CHECK_RECORDS_DOMAIN']}/check-records",
```
malcolmbaig: This isn't currently a valid route in the app — we could change it to one of these perhaps:

```suggestion
             post_logout_redirect_uri: "#{ENV['CHECK_RECORDS_DOMAIN']}/check-records/search",
```


```suggestion
             post_logout_redirect_uri: "#{ENV['CHECK_RECORDS_DOMAIN']}/check-records/sign-in",
```

Alternatively, we could add this to the Check routes (within the namespace block):
  `get "/", to: redirect("/check-records/search")`

malcolmbaig:
Not sure if I'm missing something, but when testing locally this doesn't sign me out of Check. After clicking Sign out, my session for the service is still present. Doesn't look like it reaches SignOutController#new?

config/initializers/omniauth.rb
steventux: Ah OK. I was going by what is configured in the Sign In Manage configuration (all logout redirect URLs end in `/check-records`). I had assumed these were valid paths. I'll update Manage to point to `/check-records/sign-out`.

steventux: I think this is an issue which also affects the termination of the application session.
So the redirect needs to be the sign out path on the application.
This would mean that the loop is:
Click _Sign out_
-> sign out of DfE Sign in
-> redirect to application sign out path (end of application session)
-> redirect to sign in path on app

steventux: @malcolmbaig I've amended the `post_logout_redirect_uri` value in `config/initializers/omniauth.rb` and updated the corresponding URLs in the dev Manage Sign in service.
This appears to now work as described above and invalidates the application session.

malcolmbaig: @steventux looks good, seeing a sign-in prompt from DSI after signing out of Check and then signing in again. 👍🏼

malcolmbaig (approved):
🚀

PR #377, opened 2023-10-04
Update template IDs in mailers

malcolmbaig:
This matches the template set up in the AYTQ service in Notify. https://www.notifications.service.gov.uk/services/de4ceefe-1abe-460f-9936-1991b7d26b13

The previous values are from templates labelled as AYTQ under the TRA service https://www.notifications.service.gov.uk/services/539aec1c-26ed-4f89-83c6-fb83737f3db1. We should probably use the actual AYTQ service given it's set up, as:

- this lets us generate test keys for a specific service rather than everything under TRA
- new templates don't need manually prefixing with the service name

I'll write up a card to clean up the AYTQ stuff that currently sits under the TRA service.


malcolmbaig:
Need to double check something — will reopen shortly.

PR #384, opened 2023-10-10
Update / fix DfE signin flows

steventux:
### Context

There's a CSRF issue with a user signing in, choosing an org, then signing out and signing back in again, this is due to the user session not being reset.
Similarly if a user signs in then visits the sign in path in another window they get the same error.
@malcolmbaig mentioned that a similar problem may also exist with clicking the back button after signing in.

If a user signs in to the service but does not have the appropriate role to access the service they see the _Not Authorised_ page.

- Reset the session on sign out. This allows the subsequent sign in after sign out to work as expected.
- Reset the session on sign in, this allows the user to revisit the sign in page without error, either via the back button or another window.
- Update the content of the _Not Authorised_ page adding a link which takes the user back through DSI sign out and redirects them back to the service sign in page.
- Remove the Sign in/out header navigation item for the _Not Authorised_ page. (We may revisit this but it's confusing as is). 

image

### Guidance to review

We will need to test the flows here in a non-review environment as we need real-world DSI integration.

### Link to Trello card

https://trello.com/c/mmBx9QjY/232-update-fix-dfe-signin-flows

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/auth_failures_controller.rb
+++ b/app/controllers/auth_failures_controller.rb
@@ -8,6 +8,8 @@ class AuthFailuresController < ApplicationController
     when :identity
       handle_failure_then_redirect_to qualifications_root_path
     when :dfe
+      return redirect_to(check_records_dsi_sign_out_path(id_token_hint: session[:id_token])) if session_expired?
+
       handle_failure_then_redirect_to check_records_sign_in_path(oauth_failure: true)
     end
   end
@@ -15,7 +17,6 @@ class AuthFailuresController < ApplicationController
   private
 
   def handle_failure_then_redirect_to(path)
-    error_message = request.env["omniauth.error"]&.message
     oidc_error = OpenIdConnectProtocolError.new(error_message)
     unless Rails.env.development?
       Sentry.capture_exception(oidc_error)
@@ -24,5 +25,13 @@ class AuthFailuresController < ApplicationController
     end
 
     raise oidc_error
+  end
+
+  def error_message
+    @error_message ||= request.env["omniauth.error"]&.message
+  end
+
+  def session_expired?
+    error_message == "sessionexpired"
   end
 end
--- a/app/controllers/auth_failures_controller.rb
+++ b/app/controllers/auth_failures_controller.rb
@@ -8,7 +8,7 @@ class AuthFailuresController < ApplicationController
     when :identity
       handle_failure_then_redirect_to qualifications_root_path
     when :dfe
-      handle_failure_then_redirect_to check_records_sign_in_path
+      handle_failure_then_redirect_to check_records_sign_in_path(oauth_failure: true)
     end
   end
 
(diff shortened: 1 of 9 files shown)


steventux:
[Update/Fix DfE Signin flows](https://trello.com/c/mmBx9QjY/232-update-fix-dfe-signin-flows)

gpeng (approved):

app/controllers/auth_failures_controller.rb
```diff
@@ -25,4 +26,23 @@ def handle_failure_then_redirect_to(path)
...
+    when :dfe
+      check_records_sign_in_path(oauth_failure:)
+    end
+  end
+
+  def error_message
+    @error_message ||= request.env["omniauth.error"]&.message
+  end
+
+  def sessionexpired?
```
malcolmbaig: Nit: `session_expired?`

app/controllers/auth_failures_controller.rb
```diff
@@ -8,14 +8,15 @@ def failure
     when :identity
       handle_failure_then_redirect_to qualifications_root_path
     when :dfe
-      handle_failure_then_redirect_to check_records_sign_in_path
+      return redirect_to(check_records_dsi_sign_out_path(id_token_hint: session[:id_token])) if sessionexpired?
+
+      handle_failure_then_redirect_to check_records_sign_in_path(oauth_failure: true)
```
malcolmbaig, line 13: I wonder if we can simplify here by just sending the user to the dsi_sign_out_path regardless of what the failure is? Would that still work if session[:id_token] is nil?

app/controllers/auth_failures_controller.rb
```diff
@@ -25,4 +26,23 @@ def handle_failure_then_redirect_to(path)
...
+  def redirect_path
+    strategy = request.env["omniauth.error.strategy"].name
+
+    case strategy
+    when :identity
+      qualifications_root_path
+    when :dfe
+      check_records_sign_in_path(oauth_failure:)
+    end
+  end
```
malcolmbaig: We don't seem to be using this above.

app/controllers/check_records/sign_out_controller.rb
```diff
@@ -1,6 +1,7 @@
 module CheckRecords
   class SignOutController < CheckRecordsController
     skip_before_action :handle_expired_session!
+    before_action :reset_session
 
     def new
       session[:dsi_user_id] = nil if dsi_user_signed_in?
```
malcolmbaig: Do we still need this if we `reset_session` above?

malcolmbaig (approved):
This is working nicely in Test. Left some comments, largely about cleaning up and simplifying the code.

app/controllers/auth_failures_controller.rb
steventux: I agree it would be simpler but without a token the user flow will end at an unbranded DSI page with no navigation to return to the service. At least this is the case on the dev, test and preprod sign in services.

PR #385, opened 2023-10-10
Display the restrictions in the search results

felixclack:
The prototype for the search results now displays the restrictions of
the teacher as a list.

This change introduces the design to the service.

### Link to Trello card

https://trello.com/c/85LAPLZ2/207-show-restriction-titles-on-index

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

teachers-search

### DIFF
--- a/spec/system/check_records/user_searches_for_teacher_with_restrictions_spec.rb
+++ b/spec/system/check_records/user_searches_for_teacher_with_restrictions_spec.rb
@@ -30,6 +30,8 @@ RSpec.describe "Teacher search with restrictions",
   def then_i_see_the_restriction_on_the_result
     expect(page).to have_title("Search results (1) - Check a teacher’s record")
     expect(page).to have_content("RESTRICTIONS")
+    expect(page).to have_content("Failed induction")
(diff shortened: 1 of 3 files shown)


spec/system/check_records/user_searches_for_teacher_with_restrictions_spec.rb
```diff
@@ -30,6 +30,7 @@ def and_search_returns_a_restricted_record
   def then_i_see_the_restriction_on_the_result
     expect(page).to have_title("Search results (1) - Check a teacher’s record")
     expect(page).to have_content("RESTRICTIONS")
+    expect(page).to have_content("Restrictions\t\nFailed induction\nProhibition by the General Teaching Council England (GTCE)")
```
malcolmbaig: Minor feedback: asserting on the presence of whitespace like this can be flaky, due to minor rendering variations across platforms or library versions.

An alternative might be to scope the Restrictions summary row using `within`, retrieving all the `li` tags, and then making assertions about their content.

malcolmbaig (approved):
LG — some minor feeback and a linting error aside.

PR #393, opened 2023-10-13
Encode TRNs in URLs

felixclack:
Exposing a TRN in the URL opens us up to someone being able to enumerate
through the values and capture data on people.

We want to prevent this by encoding the values passed in the URL. This
ensures that the UI is the only way to access someone's record.

I opted to use a symmetric encoding process to allow us to encode/decode
using a secret that we control.

This is an easy to understand encryption that is as secure as our secret
key.

### Link to Trello card

https://trello.com/c/RRrlB9fK/33-spike-dont-use-trn-in-the-url

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/spec/system/check_records/user_searches_with_valid_last_name_and_dob_spec.rb
+++ b/spec/system/check_records/user_searches_with_valid_last_name_and_dob_spec.rb
@@ -15,6 +15,7 @@ RSpec.describe "Teacher search", host: :check_records, type: :system do
     and_my_search_is_logged
 
     when_i_click_on_the_teacher_record
+    then_the_trn_is_not_in_the_url
     then_i_see_induction_details
     then_i_see_qts_details
     then_i_see_itt_details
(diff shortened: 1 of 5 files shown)


spec/system/check_records/user_searches_with_valid_last_name_and_dob_spec.rb
```diff
@@ -38,6 +39,10 @@ def then_i_see_a_teacher_record_in_the_results
     expect(page).to have_content "Terry Walsh"
   end
 
+  def then_the_trn_is_not_in_the_url
+    expect(page).not_to have_current_path(/teachers\/1234567/)
```
steventux: Would it be possible to be more explicit here? ie.

```ruby
expect(page).to have_current_path("/check-records/teachers/#{SecureIdentifier.encode('1234567')}")
```

steventux (approved):
This looks good to me. 
One non-blocking question about the system spec.

PR #399, opened 2023-10-23
Convert ITT and Induction qualification details fields where necessary

steventux:
### Context

The DQT API returns camelcase string values for certain fields, in the case of Induction status and ITT result fields these aren't normal human readable values. eg. `DeferredForSkillsTests`, `FailedinWales`

Use the `Hashie::Coercion` mixin to convert the values of these fields to the desired human friendly values.

### Guidance to review

I did consider using i18n inside the procs used to coerce values. 
ie. 
```ruby
 coerce_key :status, ->(value) { I18n.t("qualification.induction.status.#{value}", value.underscore.humanize) }
```

But as there were only 2 substitutions required on the Induction status values it seemed simpler to keep the coercion code in one place.
Happy to go with concensus on this though.

### Link to Trello card

https://trello.com/c/BD6V8rC0/114-ctr-present-enums-as-words

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/spec/lib/qualifications_api/teacher_spec.rb
+++ b/spec/lib/qualifications_api/teacher_spec.rb
@@ -80,6 +80,38 @@ RSpec.describe QualificationsApi::Teacher, type: :model do
       )
     end
 
+    context "ITT result field" do
+      before do
+        api_data["initialTeacherTraining"][0]["result"] = "DeferredForSkillsTests"
+      end
+
+      it "returns human readable values" do
+        expect(qualifications.find { |q| q.type == :itt }.details.result).to eq("Deferred for skills tests")
+      end
+    end
+
+    context "Induction status field" do
+      context "when the status is FailedInWales" do
+        before do
+          api_data["induction"]["status"] = "FailedInWales"
+        end
+
+        it "returns human readable values" do
+          expect(qualifications.find { |q| q.type == :induction }.details.status).to eq("Failed in Wales")
+        end
+      end
+
+      context "when the status is RequiredtoComplete" do
+        before do
+          api_data["induction"]["status"] = "RequiredtoComplete"
+        end
+
+        it "returns human readable values" do
+          expect(qualifications.find { |q| q.type == :induction }.details.status).to eq("Required to complete")
+        end
+      end
+    end
+
     context "when a qualification has no awarded date" do
       let(:api_data) do
         {
--- a/app/components/check_records/qualification_summary_component.rb
+++ b/app/components/check_records/qualification_summary_component.rb
@@ -42,8 +42,8 @@ class CheckRecords::QualificationSummaryComponent < ViewComponent::Base
 
   def itt_rows
     [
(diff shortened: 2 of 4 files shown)


steventux:
[[CTR] Present enums as words](https://trello.com/c/BD6V8rC0/114-ctr-present-enums-as-words)

spec/lib/qualifications_api/teacher_spec.rb
```diff
@@ -80,6 +80,38 @@
...
+      end
+    end
+
+    it "returns human readable Induction status values" do
+      {
+        "FailedInWales" => "Failed in Wales",
+        "InductionExtended" => "Induction extended",
+        "InProgress" => "In progress",
+        "NotYetCompleted" => "Not yet completed",
+        "PassedInWales" => "Passed in Wales",
```
malcolmbaig: Should either of these `Wales` be `wales`, to test that particular `sub` call in `CoercedDetails`?

steventux: I've gone with the specific enum values provided by the DQT API team here. 
`underscore.humanize` will convert `PassedInWales` to `Passed in wales` so the sub is acting on a lowercase `wales` here, though I agree it's not that obvious in the test. 
Could change one of these to make this more obvious?

spec/lib/qualifications_api/teacher_spec.rb
```diff
@@ -80,6 +80,38 @@
...
+        "ApplicationUnsuccessful" => "Application unsuccessful",
+        "InTraining" => "In training",
+        "NoResultSubmitted" => "No result submitted",
+        "UnderAssessment" => "Under assessment",
+      }.each do |k, v|
+        api_data["initialTeacherTraining"][0]["result"] = k
+        teacher = described_class.new(api_data)
+        qualifications = teacher.qualifications
+        expect(qualifications.find { |q| q.type == :itt }.details.result).to eq(v)
+      end
```
felixclack: All these examples appear to be the same. None of them could fail individually right?

If so, then I'd use a single example to avoid confusing things.

```suggestion
       api_data["initialTeacherTraining"][0]["result"] = "DeferredForSkillsTests"
       teacher = described_class.new(api_data)
       qualifications = teacher.qualifications
       expect(qualifications.find { |q| q.type == :itt }.details.result).to eq("Deferred for skills tests")
```

spec/lib/qualifications_api/teacher_spec.rb
felixclack: Looks like a few of these examples also test the same code path and are redundant.

steventux: I've refactored to only test the 3 different cases for value coercion.

malcolmbaig: Ah, understood 👍🏼 . The test refactor you've added since makes this clear enough I think.

malcolmbaig (approved):

PR #400, opened 2023-10-24
Give AYTQ and Check separate error controllers

malcolmbaig:
### Context

Both services use one custom error controller. This means Check has been serving up AYTQ-specific content (eg — service name, contact email.
Separate routing and controllers for each service.

### Guidance to review
Can be tested locally by navigating to `/404`, `/422`, and `/500` under each service namespace.

### Link to Trello card
https://trello.com/c/axxDTH7W

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history


app/controllers/qualifications/errors_controller.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+module Qualifications
+  class ErrorsController < QualificationsInterfaceController
+    skip_before_action :verify_authenticity_token
+    skip_before_action :authenticate_user!
+    skip_before_action :handle_expired_token!
+
+    def not_found
```
steventux: Is it worth adding these identical controller hooks/actions to a common concern?

steventux (approved):
:+1: Looks good to me, one question/suggestion about drying up the controllers, not a blocker.

gpeng (approved):

PR #411, opened 2023-11-02
Add current_namespace to ID callback controller

gpeng:
### Context

We add a namespace to the DfE Analytics web request payload to allow us to differentiate between Check and AYTQ requests. That is done in the service specific base controllers but the omniauth callback controller for AYTQ doesn't use that and so wasn't setting the namespace and consequently requests to `/qualifications/users/auth/identity/callback` aren't namespaced in BigQuery.

Set the namespace explicitly in the controller.


gpeng:
[Missing namespace - AYTQ](https://trello.com/c/5AfenRyt/146-missing-namespace-aytq)

felixclack (approved):

PR #420, opened 2023-11-13
Change how we display Mandatory Qualifications in Check

malcolmbaig:
### Context
We need to support rendering of multiple MQs, and the headings need some adjustment.

- Add support for rendering multiple MQs by following the same approach we use for NPQs
- Change the MQ section heading and make the row key more specific to accommodate for the fact that we may be rendering multiple MQs for different specialisms
- Update test data and system spec to cover these changes

Screenshot_20231113_114733

### Guidance to review
For simplicity I've put MQs just above NPQs, and below all other quals. Does ordering matter? An alternative ordering is possible but at the expense of code complexity.

### Link to Trello card
https://trello.com/c/oz1IKYZ6/144-update-mq-and-npq-rendering

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history


steventux (approved):
:100:

PR #423, opened 2023-11-13
Sort qualifications by type and date

steventux:
### Context

We’re currently combining all of the teacher qualifications and sorting by date awarded. 
This is leading to some inconsistent ordering between records.
A consistent order by qualification type, then awarded at date is preferrable.

Sorts qualifications by their type according to the typical career
progression of a teacher.
ie.
`NPQ or MQ | Induction | QTS | ITT (for QTS) | EYTS | ITT (for EYTS)`

This produces a consistent order for the types of qualification starting
with the most recent.

In the case of a specific type of qualification having multiple results,
the qualifications are sorted by reverse chronological awarded at date.

### Link to Trello card

https://trello.com/c/auIQPskM/143-update-teacher-record-row-ordering

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/app/lib/qualifications_api/teacher.rb
+++ b/app/lib/qualifications_api/teacher.rb
@@ -21,18 +21,15 @@ module QualificationsApi
     def qualifications
       @qualifications = []
 
-      add_eyts
-      add_induction
-      add_itt
-      add_mandatory_qualifications
       add_npq
+      add_mandatory_qualifications
+      add_induction
       add_qts
-      add_higher_education_qualifications
+      add_itt
+      add_eyts
+      add_itt(qts: false)
 
-      @qualifications
-        .flatten!
-        .sort_by! { |qualification| qualification.awarded_at || Date.new }
-        .reverse!
+      @qualifications.flatten!
     end
 
     def sanctions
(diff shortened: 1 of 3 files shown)


steventux:
[Update teacher record row ordering](https://trello.com/c/auIQPskM/143-update-teacher-record-row-ordering)

steventux:
[[CTR] Remove undergraduate qualifications from record](https://trello.com/c/y0SyYynU/58-ctr-remove-undergraduate-qualifications-from-record)

app/lib/qualifications_api/teacher.rb
```diff
@@ -21,18 +21,15 @@ def name
     def qualifications
       @qualifications = []
 
-      add_eyts
-      add_induction
-      add_itt
-      add_mandatory_qualifications
       add_npq
+      add_mandatory_qualifications
```
malcolmbaig, line 25: Given we can have multiple mandatory qualifications returned, do we need explicit ordering by `awarded`, as we do for NPQ?

steventux: Good spot, I hadn't noticed these could be multiples, have applied the same date sorting as others now.

malcolmbaig (approved):

PR #424, opened 2023-11-14
Add description to sanctions

felixclack:
We have expanded the sanctions list to also display a description
alongside the title.

This data isn't provided by the API so we opted to define it the same
way we do with the titles.

The descriptions contain links and we want them to be clickable so
using markdown made sense here.

The library is already being used elsewhere in the app, so there is no
extra dependency required. Also, we control the markdown text itself, so
there is no security risk by rendering this using the `html_safe`
method.

The use of `CHOMP` on the heredoc is to ensure we avoid the trailing
newline character that a heredoc automatically adds.

### Link to Trello card

https://trello.com/c/XqT7t8dV/150-add-a-description-field-for-sanctions

capybara-202311141024554670423377

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/app/models/sanction.rb
+++ b/app/models/sanction.rb
@@ -7,38 +7,225 @@ class Sanction
   end
 
   SANCTIONS = {
-    "A13" => { title: "Suspension order - with conditions" },
-    "A14" => { title: "Suspension order - with conditions" },
-    "A18" => { title: "Conditional registration order - conviction of a relevant offence" },
-    "A19" => { title: "Suspension order" },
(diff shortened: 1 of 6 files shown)


malcolmbaig (commented):
There's date-related related code in the diff that doesn't appear to be related to the PR. Approach for sanction descriptions looks sound otherwise 👍🏼

malcolmbaig (approved):

PR #432, opened 2023-11-20
Update the analytics blocklist

felixclack:
This is the output of running the regenerate task that is part of the
DfE Analytics library.


malcolmbaig:
Is there a mismatch between the output of the regenerate task and the linter?

PR #444, opened 2023-11-27
Fix display of previous names

felixclack:
The API returns the full names for previousNames and we need to reference the data correctly.

### Link to Trello card

https://trello.com/c/rluLNobI/250-250-previous-names-arent-being-correctly-displayed

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

Screenshot 2023-11-30 at 10 02 06 am

### DIFF
--- a/app/lib/qualifications_api/teacher.rb
+++ b/app/lib/qualifications_api/teacher.rb
@@ -14,10 +14,13 @@ module QualificationsApi
              :first_name,
              :last_name,
              :middle_name,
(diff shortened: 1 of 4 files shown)


gpeng (commented):
We only want to show middle and last previous names, not first_name.

felixclack:
Updated as per the discussion in standup to only display last names.

gpeng (approved):

PR #450, opened 2023-11-30
Improve tests for handling of null data from Qualifications API

malcolmbaig:
### Context

We’ve recently had a couple of production sentry errors raised because we were relying on nullable Qualifications API fields. ie. these fields were not present in the response from the API.

This PR deals specifically with how we test our handling of nullable fields. A separate PR may follow to deal with any remaining cases where we aren't correctly handling null fields, once we've completed a review.

Summary of changes:

- Do some minor tidying up of the existing `FakeQualificationsData` 
- Add a `FakeQualificationsDataWithNulling` class to return an API response with nullable fields set to nil
- Use this API response in a system spec which simply checks that the qualifications page renders without issue

### Guidance to review
I considered just copying the existing data structure in `FakeQualificationsData` and nulling fields that the [Swagger doc](https://teacher-qualifications-api.education.gov.uk/swagger/v3.json) says are nullable.

Instead I've opted for a new class (`FakeQualificationsDataWithNulling`) which takes the "full" API response and nils out specific fields. I think this makes it easier to see what the differences are between the full dataset and the nulled out one. This probably makes it easier to update in future if the API introduces any null field changes or if we discover cases we haven't accounted for.

There's some judgement exercised in deciding what to null. For example — although an entire qualification field (eg—`qts`) can be null according to the API docs, this isn't a case that's given us any issues. If there's a need to exhaustively test empty fields at this level of the data structure, we can write some more detailed unit tests for the view components that render qualifications.

There might also be a case for automatically generating the list of nullable fields (by using the Swagger doc) and applying the transformation based on this list, but I opted not to on the basis that it's a fairly heavy-handed/complex approach and the API is unlikely to change nullable properties frequently.

### Link to Trello card
https://trello.com/c/5DfQw9oj/246-check-qualifications-api-client-code-for-dependencies-on-nullable-fields

### Checklist

- [x] Attach to Trello card
- [ ] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/spec/support/fake_qualifications_api.rb
+++ b/spec/support/fake_qualifications_api.rb
@@ -1,4 +1,5 @@
 require_relative "fake_qualifications_data"
+require_relative "fake_qualifications_data_with_nulling"
 
 class FakeQualificationsApi < Sinatra::Base
   include FakeQualificationsData
@@ -11,6 +12,8 @@ class FakeQualificationsApi < Sinatra::Base
       quals_data.to_json
     when "no-itt-token"
       quals_data(trn: "1234567", itt: false).to_json
+    when "nulled-quals-data"
+      FakeQualificationsDataWithNulling.generate.to_json
     when "invalid-token"
       halt 401
     end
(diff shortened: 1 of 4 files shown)


felixclack (approved):

gpeng (approved):

steventux (approved):
LGTM! Like the nullable data approach. :+1:

PR #455, opened 2023-12-04
Separate service open feature flags

felixclack:
We have a single feature flag controlling access to 3 different apps,
which makes releasing them separately impossible.

There is no reason for them to share a single feature flag so I'm
introducing 2 extra flags to give us the granularity we want.

Each service, Check + Quals, will have a unique feature flag to control
opening the service up to the public. Also, there is now a separate
feature flag for the support interface too.

### Link to Trello card

https://trello.com/c/EiJ9iHfo/1382-add-feature-flag-to-enable-us-restrict-access-to-check-in-production-before-its-launched

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/spec/system/support/staff_sign_out_spec.rb
+++ b/spec/system/support/staff_sign_out_spec.rb
@@ -4,7 +4,8 @@ RSpec.feature "Staff sign out" do
   include CommonSteps
 
   scenario "Staff user signs out" do
-    given_the_service_is_open
+    given_the_support_service_is_open
+    given_the_qualifications_service_is_open
     and_a_staff_user_exists
     when_i_visit_the_support_interface
     then_i_see_the_staff_login_page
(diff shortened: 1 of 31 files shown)


spec/system/support/staff_sign_out_spec.rb
```diff
@@ -4,7 +4,8 @@
   include CommonSteps
 
   scenario "Staff user signs out" do
-    given_the_service_is_open
+    given_the_support_service_is_open
+    given_the_qualifications_service_is_open
```
steventux, line 8: Do we need this for a support spec?

steventux (approved):
Looks good - one question about a support spec, not a blocker.

gpeng (commented):
Looks good. Do we need to split this up so as we can set the `qualifications_service_open` flag before we start using it to avoid turning AYTQ off by mistake?

spec/system/support/staff_sign_out_spec.rb
felixclack: It's just for the redirect at the end of the spec to check that sign out works correctly.

felixclack:
> Looks good. Do we need to split this up so as we can set the `qualifications_service_open` flag before we start using it to avoid turning AYTQ off by mistake?

So deploy a version with the new qualifications_service_open flag available but not used?

PR #458, opened 2023-12-05
Add a new feature flag for AYTQ

felixclack:
We want separate feature flags controlling access to each service in this app
rather than a single one.

This will give us more granular control over releasing the apps to the
public.

As AYTQ is already a live service, we need to introduce the new flag for
it and set it to active before deploying the changes to use the new
flag.

This will allow us to seamlessly make the switch to the new names
without interruption to the live service.


felixclack:
Follow up work for this is in #455

gpeng (approved):

PR #468, opened 2023-12-07
Use db sessions

gpeng:
### Context

One of the environments (test) is throwing cookie overflow errors on the callback from DSI. Switching to ActiveRecord sessions might avoid this.

* switch to AR sessions
* install and use sidekiq-cron for keeping the sessions table tidy
* block the session data from BigQuery as we don't need it. We already record user logins via the dsi_user_sessions table.

### Checklist

- [ ] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -3,6 +3,7 @@ git_source(:github) { |repo| "https://github.com/#{repo}.git" }
 
 ruby "3.2.1"
 
+gem "activerecord-session_store"
 gem "amazing_print"
 gem "audits1984"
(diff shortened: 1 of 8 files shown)


felixclack (approved):

malcolmbaig:
Tested locally, both AYTQ and Check. Seems to be working nicely.

Suggestions

- A `Session` model would be convenient for any console-based debugging
- Might be worth deploying this to dev/test for a quick QA before merging in.

PR #481, opened 2023-12-13
Redirect root to a separate signin page

gpeng:
### Context

We are going to move the start page to GOV.UK and so don't want users to see another full start page when they land on the service.

But... the existing one is still the first result on Google so we don't want those users to get the cut down login page as they won't have seen the content on GOV.UK.

There is an existing slimmed down sign-in page that doesn't seem to be linked from anywhere so redirect to that when users land on root but leave the /qualifications/start in place until Google drops it from its index. We can redirect it then.

### Guidance to review

Screenshot 2023-12-13 at 16 56 33

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased main
- [ ] Cleaned commit history

### DIFF
--- a/app/controllers/qualifications/users/sign_in_controller.rb
+++ b/app/controllers/qualifications/users/sign_in_controller.rb
@@ -1,0 +1,11 @@
+module Qualifications
+  module Users
+    class SignInController < QualificationsInterfaceController
+      skip_before_action :authenticate_user!
+      skip_before_action :handle_expired_token!
+
+      def new
+      end
+    end
+  end
+end
--- a/spec/support/system/qualifications_authentication_steps.rb
+++ b/spec/support/system/qualifications_authentication_steps.rb
@@ -40,6 +40,6 @@ module QualificationAuthenticationSteps
   end
 
   def and_click_the_sign_in_button
-    click_button "Start now"
+    click_button "Sign in with DfE Identity"
   end
 end
--- a/spec/system/qualifications/user_visits_the_root_path_spec.rb
+++ b/spec/system/qualifications/user_visits_the_root_path_spec.rb
@@ -1,0 +1,20 @@
+require "rails_helper"
+
+RSpec.feature "The root path", type: :system do
+  include CommonSteps
+
+  scenario "User visits the root path and is directed to signin" do
+    given_the_qualifications_service_is_open
+
+    when_i_visit_the_qualifications_service
+    then_i_see_the_signin_with_identity_page
+  end
+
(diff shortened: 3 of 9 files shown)


spec/support/system/qualifications_authentication_steps.rb
```diff
@@ -40,6 +40,6 @@ def when_i_go_to_the_sign_in_page
   end
 
   def and_click_the_sign_in_button
-    click_button "Start now"
+    click_button "Sign in with Identity"
```
malcolmbaig:
```suggestion
    click_button "Sign in with DfE Identity"
```

spec/system/qualifications/user_visits_the_root_path_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+    when_i_visit_the_qualifications_service
+    then_i_see_the_signin_with_identity_page
+  end
+
+  def then_i_see_the_signin_with_identity_page
+    expect(page).to have_content(signin_button_text)
+  end
+
+  def signin_button_text
+    "Sign in with Identity"
```
malcolmbaig:
```suggestion
    "Sign in with DfE Identity"
```

app/controllers/qualifications/users/sign_in_controller.rb
```diff
@@ -0,0 +1,11 @@
+module Qualifications
+  module Users
+    class SignInController < QualificationsInterfaceController
+      skip_before_action :authenticate_user!
+      skip_before_action :handle_expired_token!
```
malcolmbaig, line 5: Suggestion: we could redirect them straight to the quals show page if there's a `current_user`.

gunndabad:
Does this affect the magic links (which link to ` /qualifications/start`)?

app/controllers/qualifications/users/sign_in_controller.rb
gpeng: Yep... agreed. I might do that in a separate PR for Jan to limit the changes we're doing this late in the day. The current start page doesn't do that so we're not making anything worse by not doing it now.

gpeng:
> Does this affect the magic links (which link to ` /qualifications/start`)?

No. It will when we remove that page in the new year but not at the mo. @johndoates is happy for us to link to the new shorter start page from those emails but we can do that in Jan. We'll redirect that path in any case so all will be well in the world.

steventux (approved):

malcolmbaig (approved):

PR #500, opened 2024-01-03
Update to the GOVUK components library 5.0

felixclack:
The design system has a new major release that contains breaking
changes.

We need to update all the relevant libraries at the same time to ensure
nothing breaks.


gpeng:
I've got a ticket in the backlog to have a look at this next sprint. We should probably hold off merging until we've launched and ironed out any post launch stuff given issues can be a bit tricky to spot in the frontend.

gpeng (approved):

PR #514, opened 2024-01-10
add tech architecture docs

aje54:
### Context

Adding tech arch docs.

updated readme.md
added architecture.md
Added aytq-ctr-container-view.dsl

https://trello.com/c/lpaDAZGq/1603-create-tech-arch-docs-for-cbl-ctr-plus-update-the-trs-doc-with-ctr


gpeng (commented):
Looks good. CTR does also make another call to DSI to get more info than comes back in the initial claim. It makes an API call to retrieve info about the user's organisation and more pertinently the role they have in that organisation context which is what we're checking for authorisation.

gpeng (approved):

PR #550, opened 2024-02-06
Fix support login configuration

steventux:
### Context

We configure a separate set of authentication routes and omniauth properties for support users.
This complexity isn't strictly necessary as we can simply examine the authenticated user's roles for an internal user role and deem them to be support users.

Refactor Omniauth config and associated code to use the same authentication callback.

### Guidance to review

This _should just work_ if you set your local `DSI_BYPASS=false`. Dev DSI config remains unchanged. 
You need to sign in via the `/check-records/sign-in` path and the policies/org/roles should take care of whether you are given access to the support UI.

### Link to Trello card

https://trello.com/c/3l2MV3KD/1758-fix-support-login-configuration

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history


steventux:
[Fix support login configuration](https://trello.com/c/3l2MV3KD/1758-fix-support-login-configuration)

app/models/dsi_user.rb
```diff
@@ -5,16 +5,19 @@ class DsiUser < ApplicationRecord
   has_many :search_logs
   has_many :dsi_user_sessions, dependent: :destroy
 
-  scope :staff, -> { where(staff: true) }
+  scope :internal, -> {
```
gpeng: Would this end up returning users as internal if they'd previously logged in as an internal user? I don't suppose it matters too much given we're only using it to list them in the support console. Maybe there's no need for that page anymore given we can no longer add or remove them inside the service. Is that page just left over from when we used devise for the staff users?

app/controllers/support_interface/support_interface_controller.rb
```diff
@@ -13,7 +13,10 @@ class SupportInterfaceController < ApplicationController
     layout "support_layout"
 
     def current_staff
```
gpeng: Do we need this anymore? Should the support interface just have it's own authorise code that checks `current_dsi_user.internal?`

app/models/dsi_user.rb
steventux: Yes the page is an artefact of devise and the various invitations and password reset functions, we could just remove all that and do away with the scope.

app/controllers/support_interface/support_interface_controller.rb
steventux: @gpeng It's used by `find_current_auditor` which is a `audits1984` requirement according to https://github.com/DFE-Digital/access-your-teaching-qualifications/commit/d98fd0c3fe1046cc009be7d3456d3c4a489a4366 and also to determine which navigation items to show. We could rename it do be less Devise-y?

app/models/dsi_user.rb
steventux: Removed Staff index here: https://github.com/DFE-Digital/access-your-teaching-qualifications/pull/550/commits/945640508b220bba6875ed4bae48976dbb6183c7

app/controllers/support_interface/support_interface_controller.rb
steventux: Removed `current_staff` here https://github.com/DFE-Digital/access-your-teaching-qualifications/pull/550/commits/da2aa4ea598e2dcdf50022a4164a636b3dcd0238

app/models/dsi_user.rb
gpeng: I don't think this scope is used any more then is it? Can we remove it altogether?

gpeng (approved):
Would be good to get @malcolmbaig to have a quick look too and test on non-prod before we merge given this is quite a sprawling change.

steventux:
I've added the new env var `DFE_SIGN_IN_API_INTERNAL_USER_ROLE_CODE` to all keyvaults so this is good to merge once @malcolmbaig has had a look and any arising comments addressed.

malcolmbaig (approved):

PR #602, opened 2024-03-11
[BUG] CTR Fix new roles authorisation bug

steventux:
### Context

A number of users who were invited to join CTR via the newly created roles have reported via Zendesk that they get the Not Authorised response.

Investigation showed that Role#internal was nil instead of false for two newly created roles, this was enough to some prevent users from successful authorisation.

- Ensure `Role#internal` is set to `false` when the form field is unchecked
- Add not null constraint to `roles.internal`

### Link to Trello card

https://trello.com/c/oCd1cgTw/1897-roleinternal-should-not-be-nil

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history


steventux:
[Role#internal should not be nil](https://trello.com/c/oCd1cgTw/1897-roleinternal-should-not-be-nil)

malcolmbaig (approved):

PR #631, opened 2024-04-17
Guard against nil previous last name

steventux:
### Context

We've seen this error a few times in production, it's simple to guard against nil API data in this case.

Sentry error(s): https://dfe-teacher-services.sentry.io/issues/5210189890

Filter out any nil last name values from a teacher's previous last names

### Checklist

- [ ] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history


richardpattinson (approved):
Simple enough change; nice job

PR #636, opened 2024-04-22
Add feedback form to AYTQ

richardpattinson:
### Context
CTR and CBL both have built in feedback forms but AYTQ is still using a Microsoft form which shold be replaced as it  and doesn’t automatically link to BigQuery.

Note the migration to add the AYTQ class isnt present yet; postgres issues on my end are preventing a migration file from being created. 

### Link to Trello card

](https://trello.com/c/9qSAiX4y/1-add-bespoke-feedback-form-for-aytq)

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased main
- [ ] Cleaned commit history

### DIFF
--- a/app/controllers/qualifications/feedbacks_controller.rb
+++ b/app/controllers/qualifications/feedbacks_controller.rb
@@ -1,7 +1,7 @@
-module CheckRecords
-  class FeedbacksController < CheckRecordsController
-    skip_before_action :authenticate_dsi_user!
-    skip_before_action :handle_expired_session!
+module Qualifications
+  class FeedbacksController < QualificationsInterfaceController
+    skip_before_action :authenticate_user!
+    skip_before_action :handle_expired_token!
 
     def new
       @feedback = Feedback.new
@@ -9,10 +9,10 @@ module CheckRecords
 
     def create
       @feedback = Feedback.new(feedback_params)
-      @feedback.service = :check
+      @feedback.service = :aytq
 
       if @feedback.save
-        redirect_to check_records_success_path
+        redirect_to qualifications_success_path
       else
         render :new
       end
--- a/app/models/feedback.rb
+++ b/app/models/feedback.rb
@@ -1,4 +1,6 @@
 class Feedback < ApplicationRecord
+  enum service: {check: "check", aytq: "aytq"}
+
   SATISFACTION_RATINGS = %w[
     very_satisfied
     satisfied
--- a/db/migrate/20240422123545_add_service_to_feedback.rb
+++ b/db/migrate/20240422123545_add_service_to_feedback.rb
@@ -1,0 +1,8 @@
+class AddServiceToFeedback < ActiveRecord::Migration[7.1]
+  def change
+    change_table :feedbacks  do |f|
+      f.column :service, :string, default: :check, null: false
+    end
+    change_column_default :feedbacks, :service, from: :check, to: nil
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[7.1].define(version: 2024_04_15_144005) do
+ActiveRecord::Schema[7.1].define(version: 2024_04_22_123545) do
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
 
@@ -126,6 +126,7 @@ ActiveRecord::Schema[7.1].define(version: 2024_04_15_144005) do
     t.string "email"
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
+    t.string "service", null: false
   end
 
   create_table "roles", force: :cascade do |t|
--- a/spec/system/qualifications/user_gives_feedback_spec.rb
+++ b/spec/system/qualifications/user_gives_feedback_spec.rb
@@ -1,0 +1,73 @@
+# frozen_string_literal: true
+require "rails_helper"
+
+RSpec.feature "Feedback", type: :system do
+  include CommonSteps
+  include QualificationAuthenticationSteps
+
+  scenario "User gives feedback", test: %i[with_stubbed_auth with_fake_quals_api] do
+    given_the_qualifications_service_is_open
+    and_i_am_signed_in_via_identity
+    and_i_click_on_feedback
+    then_i_see_the_feedback_form
+    when_i_press_send_feedback
+    then_i_see_validation_errors
+    when_i_choose_satisfied
+    when_i_fill_in_how_we_can_improve
+    when_i_choose_yes
+    when_i_enter_an_email
+    when_i_press_send_feedback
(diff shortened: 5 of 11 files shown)


app/controllers/qualifications/feedbacks_controller.rb
```diff
@@ -0,0 +1,29 @@
+class FeedbacksController < ApplicationController
```
malcolmbaig: Here we need to inherit from the `QualificationsInterfaceController`, the parent controller for most routes accessed via the AYTQ service.

app/controllers/qualifications/feedbacks_controller.rb
```diff
@@ -0,0 +1,29 @@
...
+  skip_before_action :authenticate_dsi_user!
+  skip_before_action :handle_expired_session!
```
malcolmbaig: The auth methods that need skipping are different in AYTQ - see the `QualificationsInterfaceController`.

app/models/aytq_feedback.rb
```diff
@@ -0,0 +1,14 @@
+class AYTQFeedback < ApplicationRecord
```
malcolmbaig: The Trello card suggests using the same model/table as Check. I think it's fine to just add a `service` enum to `Feedback` with values `aytq` and `check`.

spec/system/qualifications/user_gives_feedback_spec.rb
```diff
@@ -0,0 +1,73 @@
+# frozen_string_literal: true
+require "rails_helper"
+
+RSpec.feature "Feedback", host: :check_records, type: :system do
```
malcolmbaig: There's Check-specific code in this spec. It'll need going through and reworking so that it's executing against AYTQ.

malcolmbaig (changes requested):

app/controllers/qualifications/feedbacks_controller.rb
```diff
@@ -0,0 +1,31 @@
+module Qualifications
+  class FeedbacksController < QualificationsInterfaceController
+    before_action :authenticate_user!
+    before_action :handle_expired_token!
+    def new
```
malcolmbaig, line 6: whitespace before method

app/controllers/qualifications/feedbacks_controller.rb
```diff
@@ -0,0 +1,31 @@
+module Qualifications
+  class FeedbacksController < QualificationsInterfaceController
+    before_action :authenticate_user!
+    before_action :handle_expired_token!
```
malcolmbaig: These are `skip_before_action`

app/models/feedback.rb
```diff
@@ -1,4 +1,6 @@
 class Feedback < ApplicationRecord
+  enum service: {check: 0, ayts: 1}, _default: :check
```
malcolmbaig: Using symbols rather than integers would be clearer when inspecting the DB directly, IMHO.

malcolmbaig: I don't think we should specify a default - client code should be responsible for setting the appropriate value. We could pair this with a null false on the column so it fails loudly if not set.

malcolmbaig: Typo: `ayts` -> `aytq`

db/schema.rb
```diff
@@ -126,6 +126,7 @@
     t.string "email"
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
+    t.integer "service", default: 0
   end
```
malcolmbaig, line 130: This PR is missing the migration that applies these changes.

db/schema.rb
```diff
@@ -126,6 +126,7 @@
     t.string "email"
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
+    t.integer "service", default: 0
```
malcolmbaig: I think this should be a null false but with no default value. The migration could:

1. Add the column with a default value of `check` 
2. Change the column default to nil

spec/system/qualifications/user_gives_feedback_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+  include CommonSteps
+  include QualificationAuthenticationSteps
+
+  scenario "User gives feedback", test: %i[with_stubbed_auth with_fake_quals_api] do
+    given_the_qualifications_service_is_open
+    and_i_am_signed_in_via_identity
+    and_i_click_on_feedback
+    then_i_see_the_feedback_form
+    when_i_press_send_feedback
+    then_i_see_validation_errors
```
malcolmbaig, line 14: There are several invocations of `then_i_see_validation_errors` following this one. You aren't resubmitting the form, so these steps aren't testing a state change.

Suggestion: test for the validation errors once, then simply complete and submit the form.

db/migrate/20240422123545_add_service_to_feedback.rb
```diff
@@ -0,0 +1,5 @@
+class AddServiceToFeedback < ActiveRecord::Migration[7.1]
+  def change
+    add_column :feedbacks, :service, :string, :default => :check, null: false
```
malcolmbaig: As per previous comment, we should then do the following to remove the default value on this column (we're only setting it initially to backfill existing rows):

`change_column_default :feedback, :service, from: :check, to: nil`

app/models/feedback.rb
```diff
@@ -1,4 +1,6 @@
 class Feedback < ApplicationRecord
+  enum service: {check: :check, aytq: :aytq}
```
malcolmbaig: The enum values need to be strings rather than symbols, otherwise they don't get set properly on the feedback ActiveRecord object.

spec/system/qualifications/user_gives_feedback_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+    when_i_choose_satisfied
+    when_i_fill_in_how_we_can_improve
+    when_i_choose_yes
+    when_i_enter_an_email
+    when_i_press_send_feedback
+    then_i_see_the_feedback_sent_page
```
malcolmbaig, line 20: I would add a final step here (something like `and_my_feedback_is_saved`) which directly evaluates the present of a feedback record in the database, and checks that its `service` value is `aytq`.

A similar change should be made to the existing spec for Check feedback. 

This would have helped, for example, in catching bugs like https://github.com/DFE-Digital/access-your-teaching-qualifications/pull/636/files#r1582842956

malcolmbaig:
General comment: please rebase branch to remove the merge commits in the history.

db/migrate/20240422123545_add_service_to_feedback.rb
```diff
@@ -0,0 +1,8 @@
+class AddServiceToFeedback < ActiveRecord::Migration[7.1]
+  def change
+    change_table :suppliers  do |f|
```
malcolmbaig:
```suggestion
    change_table :feedbacks  do |f|
```

malcolmbaig (approved):

PR #644, opened 2024-04-30
Dont allow submitting a details change if dob pending

richardpattinson:
### Context

The account page allows users to submit a request to change the date of birth on their teaching record. If a request has already been submitted and is in-flight the user should not be allowed to submit another.

Amend the account page to show an In review tag in place of the Change link on the Date of birth row if the TRS API returns pendingDateOfBirthChange: true.

Don’t allow users to access any of the name change pages if the TRS API returns pendingDateOfBirthChange: true.

### DIFF
--- a/app/lib/qualifications_api/teacher.rb
+++ b/app/lib/qualifications_api/teacher.rb
@@ -25,6 +25,10 @@ module QualificationsApi
       previous_last_names.reject { |name| name.downcase == last_name.downcase }.map(&:titleize)
     end
 
+    def pending_dob_change?
+      api_data.pending_dob_change
+    end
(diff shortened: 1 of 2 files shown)


malcolmbaig:
As discussed, these changes are intended for the account page we're currently building for One Login users.

malcolmbaig (changes requested):
As discussed, these changes are intended for the account page we're currently building for One Login users.

malcolmbaig:
closing, see https://trello.com/c/XIQ5hsns/38-dont-allow-submitting-a-dob-change-if-there-is-already-one-pending

PR #662, opened 2024-05-15
Add request name change to One Login account page

malcolmbaig:
### Context

The AYTQ account page allows users to submit a request to change their name, along with supporting evidence.

A form that accepts name change details including an evidence upload.

PR contains complete workflow and all config changes to persist files to Azure Blob Storage via ActiveStorage.

### Guidance to review
Per-commit review highly recommended.

### Link to Trello card
https://trello.com/c/TAjAbdH5/35-change-name-page

### Checklist

- [x] Attach to Trello card
- [ ] Rebased main
- [x] Cleaned commit history


app/lib/qualifications_api/client.rb
```diff
@@ -4,14 +4,34 @@ class ForbiddenError < StandardError; end
...
+        lastName: name_change.last_name,
+        evidenceFileName: name_change.evidence_filename,
+        evidenceFileUrl: name_change.expiring_evidence_url,
+      }.to_json
+
+      response = client.post(endpoint) do |req|
+        req.body = body
+      end
+
+      response.body.fetch "caseNumber"
```
richardpattinson, line 37: Why is a method called send_name_change returning a caseNumber?

app/controllers/qualifications/one_login_users/name_changes_controller.rb
```diff
@@ -34,6 +34,18 @@ def show
         @name_change = current_user.name_changes.find(params[:id])
       end
 
+      def confirm
+        @name_change = current_user.name_changes.find(params[:id])
+        reference_number = qualifications_api_client.send_name_change(name_change: @name_change)
```
richardpattinson, line 40: Should reference number be named case number; as thats what it is in the API? Should we consider changing the name of the send name change method if its returning a case number?

richardpattinson (approved):
Some minor questions but it looks good, is there a way I could see a deployment?

app/lib/qualifications_api/client.rb
malcolmbaig: The method wraps a POST request. It sends form data for a name change and receives a case number (or an error) in response.

What alternative name would you suggest?

app/controllers/qualifications/one_login_users/name_changes_controller.rb
malcolmbaig: The TRS API calls it a case number and the AYTQ frontend calls it a reference number. There's a difference in nomenclature. I've opted to use reference number within AYTQ, treating the API client as the boundary between  AYTQ and TRS.

PR #666, opened 2024-05-17
Add guard condition to prevent accessing show page without params

richardpattinson:
### Context

Sometimes a user makes it to the show page without having date of birth params accessible. This should be prevented as it throws an unspecified error. A redirection to the search page with a notice cleanly fixes this case.

### Link to Trello card
(https://trello.com/c/Ro3rrutT/68-fix-search-params-bug-in-ctr)

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/check_records/search_controller.rb
+++ b/app/controllers/check_records/search_controller.rb
@@ -7,6 +7,13 @@ module CheckRecords
     end
 
     def show
+      unless params["search"] &&
+        params["search"].key?("date_of_birth(1i)") &&
+        params["search"].key?("date_of_birth(2i)") &&
+        params["search"].key?("date_of_birth(3i)")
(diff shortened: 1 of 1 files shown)


malcolmbaig (approved):

app/controllers/check_records/search_controller.rb
```diff
@@ -7,6 +7,11 @@ def new
     end
 
     def show
+      unless params["search"].key?("date_of_birth(1i)") &&
```
malcolmbaig: Presumably it's possible to hit this route without a `search` key set in the params? In that case, this code will raise an error.

malcolmbaig (changes requested):
One question inline. Please also confirm (via the checklist) that this has been tested locally.

malcolmbaig (approved):

PR #688, opened 2024-06-03
Refactor navigation code to fix error pages

malcolmbaig:
### Context
AYTQ 404 pages (and other HTTP error pages) aren't rendering properly because of a bug in the navigation helper. Summary:

- The error pages sit in routes without a service-specific namespace, eg - http://hostname.tld/404
- The navigation helper relies on a 'qualifications' namespace being present to render the AYTQ nav
- If this isn't present, the helper attempts to render the support nav instead
- This fails because there isn't a valid dsi_user signed in (but even if there was this would be the wrong thing to do regardless)

Fix this by doing the following:

- Extract a Qualifications::NavigationComponent out of the navigation helper method
- Update the base layout to allow the navigation to be specified in a content_for block
- Have the AYTQ layout use the new nav component, while the Support layout continues to use the helper method

This leaves us in the following state:

- AYTQ uses its own navigation component, removing the logical check and thus fixing the bug described above.
- CTR remains unchanged as it was using its own nav component already.
- Support renders its nav with a renamed helper method, although this should probably be extracted into a component at some point as well.

### Guidance to review
Can be tested locally by simply navigating to AYTQ, Check, and Support, and observing the nav they each render.

### Link to Trello card
https://trello.com/c/tGuDFTRO/70-fix-404-page-rendering

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/app/components/qualifications/navigation_component.html.erb
+++ b/app/components/qualifications/navigation_component.html.erb
@@ -1,0 +1,20 @@
+<%=
+  govuk_header(service_name: t("service.name"), service_url: qualifications_root_path) do |header|
+    if current_user
+      if FeatureFlags::FeatureFlag.active?(:one_login)
+        header.with_navigation_item(
+          active: current_page?(qualifications_identity_user_path),
+          href: qualifications_one_login_user_path,
+          text: "Account"
+        )
+      else
+        header.with_navigation_item(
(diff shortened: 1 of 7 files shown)


richardpattinson (approved):
Great work; good to ship

PR #749, opened 2024-07-17
Add better messaging for session expiry

felixclack:
When a session expires, we sign the user out and redirect them to the
start page with no explanation.

This change introduces a flash message with a small explanation in an
attempt to make the scenario clearer.

I also modified the session reset to only remove the user information
rather than clobbering the whole session. This ensures that flash
messages are persisted across redirects.


richardpattinson (approved):
Looks great to me

PR #770, opened 2024-08-07
Move the timestamp representation to a component

felixclack:
We display the timestamp of when an action was taken in a few places and
the implementation varies slightly in each spot.

I opted to move this to a ViewComponent to make it easier to test this
component in isolation.

Also, we can use the DATE_FORMATS setting to avoid hard-coding date
formats across the code and tests.

### Link to Trello card

https://trello.com/c/ICDKsFk1/245-incorrect-timestamp-on-searched-at-and-viewed-at-in-ctr

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased main
- [ ] Cleaned commit history

### DIFF
--- a/app/components/action_at_component.html.erb
+++ b/app/components/action_at_component.html.erb
@@ -1,0 +1,3 @@
+<span class="govuk-caption-m">
+  <%= action.titleize %> at <%= at %>
+</span>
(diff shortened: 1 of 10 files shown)


richardpattinson (approved):
Looks good!

PR #803, opened 2024-08-28
1901 aytq deploy review apps using workflow

shaheislamdfe:
### Context

To allow for deploy based on a specific label for review apps into new AKS environment.

Addiitonal deploy step for AKS aswell as specific Github action deploy workflow tailored to AKS deploy, referencing the aks make targets.

### Guidance to review

Add the label to the PR deploy-aks and watch review app generate with working URL based on PR number. Validate smoke tests pass also. Close PR and validate destroy action runs

### Link to Trello card

https://trello.com/c/nS67c3cG/1901-aytq-deploy-review-apps-using-workflow

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history


.github/actions/deploy-environment-aks/action.yml
```diff
@@ -0,0 +1,56 @@
...
+        PR_NUMBER: ${{ inputs.pull-request-number }}
+
+    - name: Extract Terraform outputs
+      shell: bash
+      id: set_outputs
+      run: |
+        output=$(terraform -chdir=terraform/application output -json urls)
+        echo "HOSTNAMES=$output" >> $GITHUB_ENV
+
+    - name: Run smoke tests
```
RMcVelia, line 46: why not using the existing smoke-test action?

.github/workflows/build-and-deploy.yml
```diff
@@ -69,6 +69,39 @@ jobs:
...
+
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Deploy to AKS
+        uses: ./.github/actions/deploy-environment-aks
+        id: deploy_aks
+        with:
+          environment: aks-review
+          image-tag: ${{ github.sha }}
```
RMcVelia, line 89: for consistency might be better to use ${{ needs.build_image.outputs.image_name_tag }}

.github/workflows/delete-review-app.yml
```diff
@@ -1,131 +1,77 @@
...
-          --account-key ${{ env.TFSTATE_CONTAINER_ACCESS_KEY }} \
-          --account-name ${{ env.storage_account_name }}
+      - name: Post Pull Request Comment ${{ github.event.number }}
+        if: ${{ github.event_name == 'pull_request' }}
+        uses: marocchino/sticky-pull-request-comment@v2
+        with:
+          header: aks
+          message: |
+            Review app Access Your Teaching Qualifications deployed to <https://access-your-teaching-qualifications-pr-${{ github.event.number }}.test.teacherservices.cloud> was deleted
+            Review app Check A Teacher's Record deployed to <https://check-a-teachers-record-pr-${{ github.event.number }}.test.teacherservices.cloud> was deleted
```
RMcVelia: good to keep the tf state file deletion

.github/workflows/delete-review-app.yml
```diff
@@ -1,131 +1,77 @@
...
-      - uses: Azure/login@v1
+      - name: Post Pull Request Comment
+        if: ${{ github.event_name == 'pull_request' }}
+        uses: marocchino/sticky-pull-request-comment@v2
         with:
-          creds: ${{ secrets.AZURE_CREDENTIALS }}
+          header: aks
+          message: |
+            Review app  track and pay deployed to <https://track-and-pay-${{ github.event.number }}.test.teacherservices.cloud> was deleted
+            Review app  school placements deployed to <https://manage-school-placements-${{ github.event.number }}.test.teacherservices.cloud> was deleted
```
RMcVelia: not track and pay and school placements

.github/workflows/delete-review-app.yml
```diff
@@ -1,131 +1,77 @@
...
-        description: PR number of review app to delete
-        required: true
-        type: string
+    branches:
+      - main
+    types:
+      - closed
+
 jobs:
   delete-review-app:
```
RMcVelia, line 17: if this job is for existing non-aks review apps I don't think it will work. Do we need to keep the old review apps? Maybe check with devs.

shaheislamdfe: Leaving this as suggested by Colin for now, i guess we'll have a ticket in the future for clean up and finalising the migration

.github/workflows/build-and-deploy.yml
shaheislamdfe: This is consistent with the approach thats currently set, for the standard review app deploy the image-tag: ${{ github.sha }}, if it's set to ${{ needs.build_image.outputs.image_name_tag }} then it will append the image reference incorrectly as that's used for the image_name_tag which is the first part of the reference i believe

 image_name_tag: ${{ needs.build_image.outputs.image_name_tag }}

saliceti: Eventually we should replace ./.github/actions/build-image with https://github.com/DFE-Digital/github-actions/tree/master/build-docker-image and it would output the single tag. We can keep it simple for now and use ${{ github.sha }}

.github/workflows/delete-review-app.yml
saliceti: Why delete the old code? It's required to delete the old review apps

RMcVelia: For the last few paas migrations we started removing the paas review apps as we created the aks ones, and moved across any existing. So just a question if we were do the same here. I didn't see the point of testing non-aks build/delete in this pr until that was answered.

.github/workflows/delete-review-app.yml
```diff
@@ -129,3 +133,90 @@ jobs:
           az storage blob delete -c ${{ env.tf_state_container }} --name ${{ env.TF_STATE_FILE }} \
           --account-key ${{ env.TFSTATE_CONTAINER_ACCESS_KEY }} \
           --account-name ${{ env.storage_account_name }}
+
+      - name: Post Pull Request Comment
+        if: ${{ github.event_name == 'pull_request' }}
+        uses: marocchino/sticky-pull-request-comment@v2
+        with:
+          message: |
```
saliceti: The URL is not correct for the old review apps. I suggest we don't post any comment for them and simply remove this step

saliceti (approved):

PR #855, opened 2024-09-26
Add notification banner to inform users of service outage

richardpattinson:
### Context

As part of the scheduled upgrade to AKS we will need to take the AYTQ/CTR down for an hour. This has been scheduled for monday 30th. A notification banner should be raised to notify users.

Screenshot 2024-09-26 at 21 26 23

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/app/views/layouts/base.html.erb
+++ b/app/views/layouts/base.html.erb
@@ -34,6 +34,9 @@
       <%= govuk_back_link(href: yield(:back_link_url)) if content_for?(:back_link_url) %>
       <%= yield(:breadcrumbs) if content_for?(:breadcrumbs) %>
       <main class="govuk-main-wrapper" id="main-content" role="main">
+        <% if FeatureFlags::FeatureFlag.active?(:downtime_banner) %>
+          <%= govuk_notification_banner title: "Planned downtime", text: "The service will be unavailable on Monday 30th September between 6pm and 7pm while we carry out essential maintenance" %>
+        <% end %>
         <%= render(FlashMessageComponent.new(flash: flash)) %>
         <%= yield :content %>
       </main>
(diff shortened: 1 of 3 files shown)


app/views/layouts/base.html.erb
```diff
@@ -34,6 +34,9 @@
       <%= govuk_back_link(href: yield(:back_link_url)) if content_for?(:back_link_url) %>
       <%= yield(:breadcrumbs) if content_for?(:breadcrumbs) %>
       <main class="govuk-main-wrapper" id="main-content" role="main">
+        <%= render(FlashMessageComponent.new(flash: {
+          "info" => ["Scheduled service downtime", "The service will be unavailable on Monday 30th September after 5pm while scheduled upgrades are performed"]
+        })) %>
```
felixclack: This is the wrong component to use. This isn't a flash message.

Just call the `govuk_notification_banner` method directly. See the [docs](https://govuk-components.netlify.app/components/notification-banner/#input-erb-notification-banner-with-text-arguments) for details on how to use it.

Also, we should wrap this in a feature flag so we have better control of the visibility.

```suggestion
        <% if FeatureFlags::FeatureFlag.active?(:downtime_banner)
          <%= govuk_notification_banner title: "Schedule service downtime", text: "The service will be unavailable on Monday 30th September after 5pm while scheduled upgrades are performed" %>
        <% end %>
```

felixclack (changes requested):
Left a comment about using the correct component.

app/views/layouts/base.html.erb
```diff
@@ -34,6 +34,8 @@
       <%= govuk_back_link(href: yield(:back_link_url)) if content_for?(:back_link_url) %>
       <%= yield(:breadcrumbs) if content_for?(:breadcrumbs) %>
       <main class="govuk-main-wrapper" id="main-content" role="main">
+        <% if FeatureFlags::FeatureFlag.active?(:downtime_banner) %>
+          <%= govuk_notification_banner title: "Planned downtime", text: "The service will be unavailable on Monday 30th September between 6pm and 7pm while we carry out essential maintenance" %>        <% end %>
```
felixclack: Nit: The `end` should be on it's own line.

richardpattinson: Done

felixclack (approved):

PR #944, opened 2025-04-25
Check alert exposable and has an end date

richardpattinson:
### Context

Alerts were rendering that we do not expose in CTR. This was showing restrictions tags in places they were not appropriate. 

Add alert_type_ids for 5 alert types we should be serving. 

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased main
- [ ] Cleaned commit history


peteryates (approved):
Code wise, looks fine to me

gunndabad (approved):

PR #972, opened 2025-07-08
Update to use api-version 20250627 of v3/persons/#{trn}

Lockyy:
### Context

The API providing info powering CTR and AYTQ is being updated to use routes rather than specifically returning Initial Teacher Training info. We need to update the system to consume this information and render it out to the user.

- Updates api version for v3/persons/#{trn} to 20250627
- Swap interface to rendering routes information to users of both CTR and AYTQ

### Link to Trello card

- https://trello.com/c/2xip7gQR
- https://trello.com/c/YHFVdDXI

### DIFF
--- a/app/components/check_records/qualification_summary_component.rb
+++ b/app/components/check_records/qualification_summary_component.rb
@@ -27,23 +27,24 @@ class CheckRecords::QualificationSummaryComponent < ViewComponent::Base
   alias_method :title, :name
 
   def rows
-    @rows = (
-      if itt?
-        itt_rows
-      elsif mq?
-        mq_rows
-      elsif induction?
-        induction_rows
-      elsif qts?
-        qts_rows
-      else
-        [
-          { key: { text: "Date awarded" }, value: { text: awarded_at&.to_fs(:long_uk) } },
-          { key: { text: "Status" }, value: { text: status_description } },
-        ]
-      end
-    )
-    @rows.select { |row| row[:value][:text].present? }
+    @rows ||= build_rows.select { |row| row[:value][:text].present? }
+  end
+
+  def build_rows
+    if itt?
+      itt_rows
+    elsif mq?
+      mq_rows
+    elsif induction?
+      induction_rows
+    elsif qts?
+      qts_rows
+    else
+      [
+        { key: { text: "Date awarded" }, value: { text: awarded_at&.to_fs(:long_uk) } },
+        { key: { text: "Status" }, value: { text: status_description } },
+      ]
+    end
   end
 
   def induction_rows
--- a/app/components/check_records/qualification_summary_component.rb
+++ b/app/components/check_records/qualification_summary_component.rb
@@ -25,8 +25,7 @@ class CheckRecords::QualificationSummaryComponent < ViewComponent::Base
            to: :qualification
 
   def title
-    return "Professional Status" if qts? || eyts?
-    return "Route to Professional Status" if rtps?
+    return [name, rtps_route_type].compact.join(": ") if rtps?
 
     name
   end
@@ -77,7 +76,6 @@ class CheckRecords::QualificationSummaryComponent < ViewComponent::Base
 
   def rtps_rows
     [
-      { key: { text: "Route Type" }, value: { text: details.route_to_professional_status_type&.name } },
       { key: { text: "Qualification" }, value: { text: details.degree_type&.name } },
       { key: { text: "Provider" }, value: { text: details.training_provider&.name } },
       {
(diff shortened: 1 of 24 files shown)


app/lib/qualifications_api/teacher.rb
```diff
@@ -42,17 +44,13 @@ def previous_names
...
-      add_itt(qts: false)
-
-      @qualifications.flatten!
+      @qualifications ||= [
+        npq_qualifications,
+        mq_qualifications,
+        induction_qualification,
+        qts_qualifications,
+        eyts_qualifications,
+      ].flatten
```
Lockyy: This was rebuilding all the qualifications every time qualifications was called. I've refactored it so that it builds them all once and then memoizes the result.

I've also refactored this so that rather than each method adding to @qualifications all the assignment just happens here so that the qualifications method is the only place that knows about @qualifications, simplifying the logic of the class as a whole.

app/lib/qualifications_api/teacher.rb
```diff
@@ -16,6 +16,8 @@ class Teacher
       NPQSENCO: "National Professional Qualification (NPQ) for Special Educational Needs Co-ordinators"
     }.freeze
 
+    QTLS_ROUTE_ID = "BE6EAF8C-92DD-4EFF-AAD3-1C89C4BEC18C".freeze
```
Lockyy, line 19: ID provided by @gunndabad

app/lib/qualifications_api/teacher.rb
Lockyy: As a part of this rather than replacing the add_itt methods with equivalent methods for the new rtps system, the qts_qualificaitons and eyts_qualifications methods each return their respective rtps qualification objects so that they get inserted into the right place in the list automatically.

app/components/check_records/qualification_summary_component.rb
```diff
@@ -10,108 +10,133 @@ class CheckRecords::QualificationSummaryComponent < ViewComponent::Base
...
            :qts_and_qtls,
            to: :qualification
 
-  alias_method :title, :name
+  def title
+    return "Professional Status" if qts? || eyts?
+    return "Route to Professional Status" if rtps?
+
+    name
+  end
```
Lockyy, line 31: Previously designs showed the name of the qualification as the title, this has changed so that now that goes in the table, so I've replaced the alias with this method.

app/components/check_records/qualification_summary_component.rb
```diff
@@ -10,108 +10,133 @@ class CheckRecords::QualificationSummaryComponent < ViewComponent::Base
...
       },
       { key: { text: "Date completed" }, value: { text: awarded_at&.to_fs(:long_uk) } }
     ]
   end
 
-  def itt_rows
+  def rtps_rows
+    age_range = if details.training_age_specialism
+                  "#{details.training_age_specialism.from} to #{details.training_age_specialism.to} years"
+                end
```
Lockyy: @gunndabad The swagger docs indicate that `details. training_age_specialism.type` can equal `Range, FoundationStage, FurtherEducation, KeyStage1, KeyStage2, KeyStage3, KeyStage4`. I've not got an example of what options other than Range do to the from and to attributes. So need clarification as to if this is the correct approach.

app/models/qualification.rb
```diff
@@ -1,8 +1,17 @@
...
+  attr_accessor :awarded_at,
+                :name,
+                :type,
+                :qtls_only,
+                :set_membership_active,
+                :set_membership_expired,
+                :passed_induction,
+                :failed_induction,
+                :qts_and_qtls,
+                :routes
```
Lockyy, line 13: removed status_description
added routes

I've separated this up onto new lines so future changes are more visible in the commit history.

app/lib/qualifications_api/certificate.rb
```diff
@@ -2,7 +2,7 @@ module QualificationsApi
   class Certificate
     include ActiveModel::Model
 
-    VALID_TYPES = %i[eyts induction itt NPQEL NPQLTD NPQLT NPQH NPQML NPQLL NPQEYL NPQSL NPQLBC NPQSENCO NPQLPM 
+    VALID_TYPES = %i[eyts induction rtps NPQEL NPQLTD NPQLT NPQH NPQML NPQLL NPQEYL NPQSL NPQLBC NPQSENCO NPQLPM
```
Lockyy, line 5: This likely isn't going to work and needs revisiting.

cc @richardpattinson

app/components/check_records/qualification_summary_component.rb
gunndabad: Only `Range` will have a `to` and `from`; the others will need text that's derived from the `type`.

gunndabad (approved):

